PR #16, opened 2021-09-16
feat: add class based tooling for filters

mariajgrimaldi:
**Description:** 
This PR  adds class-based tooling for filter execution.

**Dependencies:** 
None for now.

**Merge deadline:** 
None.

**Installation instructions:** 
Install openedx-filters: `pip install git+https://github.com/eduNEXT/openedx-filters.git@MJG/class_filter_tooling#egg=openedx_filters==0.2.0_alpha`

**Testing instructions:**
1. In your edx-platform version, you can check out to this branch or add the changes yourself:
https://github.com/edx/edx-platform/pull/28773
2. Change your settings according to your test plugin and add the following:
```
from openedx_filters import PipelineStep
from openedx_filters.learning.enrollment import PreEnrollmentFilter

class CanPerformEnrollment(PipelineStep):
    """
    Filter step used as pre-enrollment approval.
    """

    def do_filter(self, user, course_key, mode):
       "Method that raises exception used to avoid enrollments."
        raise PreEnrollmentFilter.PreventEnrollment
```
3. Enroll in a  course

**Reviewers:**
- [x] @felipemontoya 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** 
None for now


openedx_filters/tooling.py
```diff
@@ -0,0 +1,231 @@
...
+        functions = cls.get_functions_for_pipeline(pipeline)
+
+        accumulated_output = kwargs.copy()
+        for function in functions:
+            try:
+                step_result = function(*args, **accumulated_output) or {}
+
+                if not isinstance(step_result, dict):
+                    log.info(
+                        "Pipeline stopped by '%s' for returning an object.",
```
felipemontoya, line 207:
```suggestion
                        "Pipeline stopped by '%s' for returning an object.",
```
```suggestion
                        "Pipeline stopped by '%s' for returning an object instead of a dict.",
```

felipemontoya (changes requested):
Also, the discussion we had about extra metadata passed to the filter function

openedx_filters/tooling.py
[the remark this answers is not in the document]
mariajgrimaldi: Some time ago, we talked about having partial pipelines, so returning an object doesn't mean the filter is broken. So, for example, if X, then I want to stop the execution. If Y, I want to continue. To stop the execution, I'd use a partial pipeline

openedx_filters/tooling.py
```diff
@@ -0,0 +1,231 @@
...
+                        function.__name__,
+                    )
+                    return step_result
+                accumulated_output.update(step_result)
+            except OpenEdxFilterException as exception:
+                if raise_exception:
```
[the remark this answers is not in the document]
mariajgrimaldi: Already considered :+1: thank you!

openedx_filters/tooling.py
```diff
@@ -0,0 +1,231 @@
...
+                    'fail_silently': False,
+                    'log_level': "debug"
+                }
+
+                Where:
+                    - pipeline (list): paths where the functions to be executed by
+                    the pipeline are defined.
+                    - fail_silently (bool): determines whether the pipeline can
+                    raise exceptions while executing. If its value is True then
+                    exceptions (OpenEdxFilterException) are caught and the execution
```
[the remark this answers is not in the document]
mariajgrimaldi: Thanks for the explanation! I updated the docstring

openedx_filters/filters.py
```diff
@@ -0,0 +1,31 @@
...
+        Arguments:
+            filter_type (str): name of the filter.
+            running_pipeline ([type]): list of steps currently running.
+            extra_config (dict): extra configuration defined in OPEN_EDX_FILTERS_CONFIG.
+        """
+        self.filter_type = filter_type
+        self.running_pipeline = running_pipeline
+        self.extra_config = extra_config
+
+    def do_filter(self, **kwargs):
```
felipemontoya, line 24: `run` ?

openedx_filters/tooling.py
```diff
@@ -0,0 +1,239 @@
...
+            except OpenEdxFilterException as exc:
+                log.exception(
+                    "Exception raised while running '%s':\n %s", step.__name__, exc,
+                )
+                raise
+            except Exception as exc:
+                # We're catching this because we don't want the core to blow up
+                # when a filter is broken and fail_silently is False. Otherwise,
+                # the pipeline resumes its execution.
+                if fail_silently:
```
felipemontoya, line 229: Should we at least log it here before being so silent?

felipemontoya (changes requested):

openedx_filters/tooling.py
```diff
@@ -0,0 +1,238 @@
...
+            **extra_config,
+        }
+
+        accumulated_output = kwargs.copy()
+        for step in steps:
+            try:
+                step_runner = step(**filter_metadata)
+                result = step_runner.run(**accumulated_output)
+
+                if not isinstance(result, dict):
```
felipemontoya, line 213: This is very strict.
If the Step returns something different than a dict, then it stops.

Does that mean that if I want to return nothing, but allow other Steps to continue I should return `{}` ?

openedx_filters/tooling.py
```diff
@@ -0,0 +1,238 @@
...
+            "filter_type": cls.filter_type,
+            "running_pipeline": pipeline,
+            **extra_config,
+        }
+
+        accumulated_output = kwargs.copy()
+        for step in steps:
+            try:
+                step_runner = step(**filter_metadata)
+                result = step_runner.run(**accumulated_output)
```
felipemontoya, line 211: during this run we are ignoring the configured 'log_level' for the pipeline. This is what will come in PR#17, right?

mariajgrimaldi: Right!

openedx_filters/tooling.py
mariajgrimaldi: Yes, or the arguments that the filter received. We inherited this behavior from the social core pipeline.

felipemontoya (approved):

PR #20, opened 2021-11-25
feat: add 1st batch of filters

mariajgrimaldi:
**Description:** 

This PR adds:

* PreRegisterFilter: intended to be executed before the registration process starts.
* PreLoginFilter: intended to be executed before the login process starts.

Used by edx-platform.

**Testing instructions:**

1. Install openedx-filters using this branch.
2. Use this edx-platform branch [MJG/1st_filters_batch](https://github.com/eduNEXT/edx-platform/tree/MJG/1st_filters_batch)
3. In your favorite plugin, create some filter steps. For example:
```
class ModifyUsernameBeforeRegistration(PipelineStep):

    def run_filter(self, form_data):
        username = f"{form_data.get('username')}-edunext"
        form_data["username"] = username
        return {
            "form_data": form_data,
        }
```
These sample filter steps are implemented in [this](https://github.com/eduNEXT/openedx-filters-samples/blob/master/openedx_filters_samples/samples/pipeline.py) plugin

4. Add the following Django settings to your configuration file:
```
OPEN_EDX_FILTERS_CONFIG = {
    "org.openedx.learning.course.enrollment.started.v1": {
        "fail_silently": False,
        "pipeline": [
            "openedx_filters_samples.samples.pipeline.ModifyModeBeforeEnrollment",
            "openedx_filters_samples.samples.pipeline.StopEnrollment"
        ]
    },
    "org.openedx.learning.student.login.requested.v1": {
        "fail_silently": False,
        "pipeline": [
            "openedx_filters_samples.samples.pipeline.ModifyUserProfileBeforeLogin",
            "openedx_filters_samples.samples.pipeline.StopLogin"
        ]
    },
    "org.openedx.learning.student.registration.requested.v1": {
        "fail_silently": False,
        "pipeline": [
            "openedx_filters_samples.samples.pipeline.ModifyUsernameBeforeRegistration",
            "openedx_filters_samples.samples.pipeline.StopRegister"
        ]
    }
}
```
With this configuration, you won't be able to register, login, or enroll in a course. Another configuration you could try is, instead of `Stop<process>` use `NoopFilter`

**Reviewers:**
- [ ] @felipemontoya 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


openedx_filters/learning/auth.py
```diff
@@ -0,0 +1,54 @@
+"""
```
mariajgrimaldi, line 1: should we move all filters to `filters.py` file? @felipemontoya to be compliant with openedx-events

openedx_filters/learning/auth.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class PreRegisterFilter(OpenEdxPublicFilter):
+    """
+    Custom class used to create PreRegister filters.
+    """
+
+    filter_type = "org.openedx.learning.student.registration.requested.v1"
+    sensitive_form_data = [
+        "password",
```
felipemontoya, line 15: Is `password` enough? should we be censoring all the fields at https://github.com/edx/edx-platform/blob/master/common/djangoapps/track/middleware.py#L64

openedx_filters/learning/auth.py
```diff
@@ -0,0 +1,89 @@
...
+    @classmethod
+    def run(cls, form_data):
+        """
+        Execute a filter with the signature specified.
+
+        Arguments:
+            form_data (QueryDict): contains the request.data submitted by the registration
+            form.
+        """
+        sensitive_data = cls.remove_sensitive_form_data(form_data)
```
felipemontoya, line 32: This should not be a blocking issue, but is this something that should be handled at the core before we even send the call to the filter?
Could it be that we end up forgetting that sensitive data is passed to the filter class and eventually end up logging it?

openedx_filters/learning/auth.py
```diff
@@ -0,0 +1,89 @@
...
+        base_form_data = form_data.copy()
+        for key, value in base_form_data.items():
+            if key in cls.sensitive_form_data:
+                form_data.pop(key)
+                sensitive_data[key] = value
+
+        return sensitive_data
+
+
+class PreLoginFilter(OpenEdxPublicFilter):
```
felipemontoya, line 62: Much the same as the naming of the filter above, is PreLogin the name or should we go with StudentLoginFilter?

felipemontoya (changes requested):

openedx_filters/learning/auth.py
```diff
@@ -0,0 +1,89 @@
+"""
+Package where filters related to the auth process are implemented.
+"""
+from openedx_filters.exceptions import OpenEdxFilterException
+from openedx_filters.tooling import OpenEdxPublicFilter
+
+
+class PreRegisterFilter(OpenEdxPublicFilter):
```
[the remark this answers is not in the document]
felipemontoya: This probably also affects the naming of the PreEnrollmentFilter and the implementing PR as well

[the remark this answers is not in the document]
mariajgrimaldi: Great suggestion. I'll go with the `StudentRegistrationRequested` because it's the closest to the events naming convention.

openedx_filters/learning/auth.py
mariajgrimaldi: Handling sensitive data within the filter implementation was Dave's recommendation in [PR 29449](https://github.com/edx/edx-platform/pull/29449#discussion_r761602507). And as I see it, the filter implementation should decide what to pass on the pipeline -in its context of login, registration, etc.- instead of relying on the core, this way we encapsulate implementation details inside the filter. What do you think?

felipemontoya: This is good then. I had not seen Dave's comment on the other PR yet.

openedx_filters/learning/filters.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Custom class used to create PreRegister filters.
+    """
+
+    filter_type = "org.openedx.learning.student.registration.requested.v1"
+    sensitive_form_data = [
+        "password", "newpassword", "new_password", "oldpassword", "old_password", "new_password1", "new_password2",
+    ]
+
+    class PreventRegister(OpenEdxFilterException):
```
felipemontoya, line 19: Nit: "preventRegistration" ?

felipemontoya (approved):
This is looking almost ready to me now. I left a comment about the name of the exception but once that is in, I''m good with merging.

PR #21, opened 2021-11-26
fix: remove "OpenEdxFilterException:" from exception str repr

mariajgrimaldi:
**Description**
This PR changes the string representation for filters exceptions.

Before
image
```
  File "/edx/src/openedx-filters/openedx_filters/tooling.py", line 211, in run_pipeline
    result = step_runner.run(**accumulated_output)
  File "/edx/src/openedx-filters-test-plugin/openedx_filters_test_plugin/pipeline.py", line 146, in run
    raise PreLoginFilter.PreventLogin("You can't login on this site.")
openedx_filters.learning.auth.PreLoginFilter.PreventLogin: OpenEdxFilterException: You can't login on this site.
```
After
image
```
  File "/edx/src/openedx-filters/openedx_filters/tooling.py", line 211, in run_pipeline
    result = step_runner.run(**accumulated_output)
  File "/edx/src/openedx-filters-test-plugin/openedx_filters_test_plugin/pipeline.py", line 146, in run
    raise PreLoginFilter.PreventLogin("You can't login on this site.")
openedx_filters.learning.auth.PreLoginFilter.PreventLogin: You can't login on this site.
```

### DIFF
--- a/openedx_filters/exceptions.py
+++ b/openedx_filters/exceptions.py
@@ -35,4 +35,4 @@ class OpenEdxFilterException(Exception):
         """
         Show string representation of OpenEdxFilterException using its message.
         """
-        return "OpenEdxFilterException: {}".format(self.message)
+        return self.message


felipemontoya:
Is this coming from this __str__ method, or is it because we are rendering `str(exception)` instead of `exception.message`?

What is the most common practice?

Also, if we are changing the `__str__` to be final-user readable, should we add a `__repr__ ` to be more developer-user readable?

mariajgrimaldi:
We are rendering str(e) @felipemontoya. On edx-platform:
- `str(exc)` is frequent than `exc.message` (~60 search results vs 1) 
- `str(e)` is frequent than `e.message` (~24 search results vs 0)

I think `str(e)` is frequent because `e.message` doesn't work with all exceptions. We can use str(e) as the user-readable message, and use the traceback as part of the  __repr__. What do you think?

felipemontoya:
So it's almost always `str(exception)`. Ok, lets go ahead with this then.

felipemontoya (approved):

PR #30, opened 2022-04-05
refactor: change course home filter exception name so it's more descriptive

mariajgrimaldi:
**Description:** 
 This PR changes the course home filter exception name so it's more descriptive


mariajgrimaldi:
Closing since we're removing the course home filter.

PR #31, opened 2022-04-05
refactor!: refactor due to integration feedback

mariajgrimaldi:
This PR:

* Remove CourseHomeRenderStarted since it's not going to be used
* Change RenderAlternativeCertificate to RenderAlternativeInvalidCertificate

### DIFF
--- a/openedx_filters/tests/test_tooling.py
+++ b/openedx_filters/tests/test_tooling.py
@@ -22,7 +22,7 @@ class FirstPipelineStep(PipelineStep):
     """
 
     def run_filter(self, **kwargs):
-        pass
+        return {}
 
 
 class SecondPipelineStep(PipelineStep):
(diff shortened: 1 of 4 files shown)


felipemontoya:
This is mostly good, but you need to fix tests and references to deleted code

mariajgrimaldi:
Well. I think we have an issue.

I understand we should increase/leave untouched the coverage with every PR, but I'm removing code -including tests that now don't make any sense- and the coverage decreases and the check fails? Am I missing something? :thinking:

openedx_filters/tests/test_tooling.py
```diff
@@ -286,6 +286,30 @@ def test_run_empty_pipeline(self):
...
+        This method runs noop pipeline, i.e, a pipeline with functions that
+        don't affect the input.
+
+        Expected behavior:
+            Returns the same input arguments.
+        """
+        result = PreEnrollmentFilterMock.run_pipeline(**self.kwargs)
+
+        self.assertDictEqual(result, self.kwargs)
+
```
mariajgrimaldi, line 312: I'm adding this test to comply with the coverage checks. Sorry for the mixup.

felipemontoya (approved):

PR #32, opened 2022-04-06
test: add tests cases to improve coverage

mariajgrimaldi:
**Description:** 
Add tests cases to improve coverage, so #31 and future build passes.


felipemontoya (approved):
100% coverage! Yeeiii

PR #42, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42


openedx_filters/filters.py
```diff
@@ -55,6 +56,7 @@ def __init__(self, filter_type, running_pipeline, **extra_config):
         self.running_pipeline = running_pipeline
         self.extra_config = extra_config
 
+    @abstractmethod
```
sarina, line 59: I added this to get around the new pylint error `R0201: Method could be a function (no-self-use)`. This feels to me like the best solution given your docstring!

sarina:
@mariajgrimaldi - this is ready for your review 😁

openedx_filters/filters.py
mariajgrimaldi: Good catch! Thank you :smile:

mariajgrimaldi (approved):
Thanks :sun_with_face:

PR #106, opened 2023-07-25
feat: added filter for modifying the isStarted for externally hosted …

jajjibhai008:
**Description**: 
This PR adds a filter for modifying the `isStarted` for externally hosted courses. This hook will identify whether the learner has started the course or not.

**Reference**: 
Actually, this field will be used to identify whether the learner has started the course or not. And we will conditionally render some things on the learner dashboard (B2C side) as described in [ENT-7188](https://2u-internal.atlassian.net/browse/ENT-7188).

**JIRA**- [ENT-7373](https://2u-internal.atlassian.net/browse/ENT-7373) and [ENT-at 7374](https://2u-internal.atlassian.net/browse/ENT-7374)

**Dependencies:** dependencies on other outstanding PRs, issues, etc. 

**Merge deadline:** List merge deadline (if any)

**Installation instructions:** List any non-trivial installation 
instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happened instead - check failed.

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** List any concerns about this PR - inelegant 
solutions, hacks, quick-and-dirty implementations, concerns about 
migrations, etc.

### DIFF
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -591,6 +591,26 @@ class CourseHomeUrlCreationStarted(OpenEdxPublicFilter):
         return data.get("course_key"), data.get("course_home_url")
 
 
+class CourseEnrollmentAPIRenderStarted(OpenEdxPublicFilter):
+    """
+    Custom class used to create filters for enrollment data.
+    """
+
+    filter_type = "org.openedx.learning.home.enrollment.api.rendered.v1"
+
+    @classmethod
+    def run_filter(cls, course_key, serialized_enrollment):
+        """
+        Execute a filter with the specified signature.
+
+        Arguments:
+            course_key (CourseKey): The course key for which isStarted is being modify.
+            serialized_enrollment (dict): enrollment data
+        """
+        data = super().run_pipeline(course_key=course_key, serialized_enrollment=serialized_enrollment)
+        return data.get("course_key"), data.get("serialized_enrollment")
+
+
 class InstructorDashboardRenderStarted(OpenEdxPublicFilter):
     """
     Custom class used to create instructor dashboard filters and its custom methods.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 ----------
 
+[1.5.0] - 2023-07-19
+--------------------
+
+Added
+~~~~~
+
+* CourseEnrollmentAPIRenderStarted filter added which can be used to modify the isStarted B2C dashboard rendering process.
+
+
 [1.4.0] - 2023-07-18
 --------------------
 
--- a/openedx_filters/__init__.py
+++ b/openedx_filters/__init__.py
@@ -3,4 +3,4 @@ Filters of the Open edX platform.
 """
 from openedx_filters.filters import *
 
-__version__ = "1.4.0"
+__version__ = "1.5.0"
--- a/openedx_filters/learning/tests/test_filters.py
+++ b/openedx_filters/learning/tests/test_filters.py
@@ -13,6 +13,7 @@ from openedx_filters.learning.filters import (
     CohortAssignmentRequested,
     CohortChangeRequested,
     CourseAboutRenderStarted,
+    CourseEnrollmentAPIRenderStarted,
     CourseEnrollmentQuerysetRequested,
     CourseEnrollmentStarted,
     CourseHomeUrlCreationStarted,
@@ -604,3 +605,17 @@ class TestFederatedContentFilters(TestCase):
         result = CourseHomeUrlCreationStarted.run_filter(course_key, course_home_url)
 
         self.assertTupleEqual((course_key, course_home_url,), result)
+
+    def test_course_enrollment_api_render_started(self):
+        """
+        Test CourseEnrollmentAPIRenderStarted filter behavior under normal conditions.
+
+        Expected behavior:
+            - The filter must have the signature specified.
+            - The filter should return course_key and is_started in that order.
+        """
+        course_key, serialized_enrollment = Mock(), Mock()
+
+        result = CourseEnrollmentAPIRenderStarted.run_filter(course_key, serialized_enrollment)
+
+        self.assertTupleEqual((course_key, serialized_enrollment,), result)


jajjibhai008:
@felipemontoya could you please review my PR? I have to ship these changes as soon as possible.

felipemontoya:
Hi @jajjibhai008 , thanks for tagging me.

Do you have the link to the corresponding edx-platform pr? I am not understanding the goal of this filter by looking at the definition alone.

muhammad-ammar:
Hi @felipemontoya, Here is the edx-platform PR: https://github.com/openedx/edx-platform/pull/32788

felipemontoya:
Thanks a lot. Now I understand much better.

I think the definition of the filter is too specific to make it a reusable filter.
Also it makes the naming difficult.

You have:
"org.openedx.learning.course.is.started.creation.started.v1"

But in order to truly pinpoint the filtered situation it would have to be:
"org.openedx.learning.home.enrollment.api.render.is.started.calculated.v1"

I think you could refactor this a bit to make it global to the whole API call.

Broadly speaking you could let the serializer run its whole logic and then add a filter that lets you work on the final result.

```python
class EnrollmentSerializer(serializers.Serializer):

    def to_representation(self, instance):
        """Let the filter work after the whole instance is serialized"""
        ret = super().to_representation(instance)
        ret = CourseEnrollmentAPIRendered().run_filter(representation=ret)
        return ret
```

So you could have a filter whose domain is:
"org.openedx.learning.home.enrollment.api.rendered.v1"

Such a filter could be re used in the future when you want to alter any of the fields of this API

felipemontoya:
Now that I think of it, maybe you would rather filter the instance object before is passed to `to_representation`. In that case you are working with actual models instead of text/json in the filter. But the overall idea stays the same.

felipemontoya (changes requested):
as per the comment above

jajjibhai008:
> Thanks a lot. Now I understand much better.
> 
> I think the definition of the filter is too specific to make it a reusable filter. Also it makes the naming difficult.
> 
> You have: "org.openedx.learning.course.is.started.creation.started.v1"
> 
> But in order to truly pinpoint the filtered situation it would have to be: "org.openedx.learning.home.enrollment.api.render.is.started.calculated.v1"
> 
> I think you could refactor this a bit to make it global to the whole API call.
> 
> Broadly speaking you could let the serializer run its whole logic and then add a filter that lets you work on the final result.
> 
> ```python
> class EnrollmentSerializer(serializers.Serializer):
> 
>     def to_representation(self, instance):
>         """Let the filter work after the whole instance is serialized"""
>         ret = super().to_representation(instance)
>         ret = CourseEnrollmentAPIRendered().run_filter(representation=ret)
>         return ret
> ```
> 
> So you could have a filter whose domain is: "org.openedx.learning.home.enrollment.api.rendered.v1"
> 
> Such a filter could be re used in the future when you want to alter any of the fields of this API

HI @felipemontoya Thanks for your detailed review.

1) I think your suggested filter name is more concise and descriptive, So I have renamed the filter name in all PRs.
2) I agree with your approach to writing the filtration logic in the `to_representation(self, instance)` method. This will make code more efficient and readable. I have update the[ platform PR](https://github.com/openedx/edx-platform/pull/32788) according to your suggestions. 
Please let me know if you have any questions or confusion.

felipemontoya (changes requested):
hey @jajjibhai008, I'm happy that the review helped.

I'd still like to understand something. You changed the position of the filter to a position where you can have an effect on [all the fields](https://github.com/openedx/edx-platform/pull/32788/files#diff-bdf2cf3efb3de96e4fe1df13d5f90e685dc7d4c0e3c3fd61c4e7424e2feb8abcL192-L203) for this API. But you are still defining the filter in a way that would only let a developer using the filter change the `is_started` value. I would still suggest that you change the definition so that by opening the capability to alter the API you are actually allowing the complete set of fields to be worked on.

Then you could call the filter `"org.openedx.learning.home.enrollment.api.rendered.v1"` and allow other features in the future to build on top of this.

jajjibhai008:
> hey @jajjibhai008, I'm happy that the review helped.
> 
> I'd still like to understand something. You changed the position of the filter to a position where you can have an effect on [all the fields](https://github.com/openedx/edx-platform/pull/32788/files#diff-bdf2cf3efb3de96e4fe1df13d5f90e685dc7d4c0e3c3fd61c4e7424e2feb8abcL192-L203) for this API. But you are still defining the filter in a way that would only let a developer using the filter change the `is_started` value. I would still suggest that you change the definition so that by opening the capability to alter the API you are actually allowing the complete set of fields to be worked on.
> 
> Then you could call the filter `"org.openedx.learning.home.enrollment.api.rendered.v1"` and allow other features in the future to build on top of this.


@felipemontoya I have updated the filter definition to `"org.openedx.learning.home.enrollment.api.rendered.v1"`.

openedx_filters/learning/filters.py
```diff
@@ -591,6 +591,26 @@ def run_filter(cls, course_key, course_home_url):
         return data.get("course_key"), data.get("course_home_url")
 
 
+class CourseEnrollmentRenderStarted(OpenEdxPublicFilter):
```
felipemontoya, line 594: If you decide to change like the comment below, I would also ask to please change the name to `CourseEnrollmentAPIRenderStarted` so that it is clear that the render is at an API.

felipemontoya (changes requested):

openedx_filters/learning/filters.py
jajjibhai008: @felipemontoya i have updated the `to_representation(self, instance)` of enrollment_serializer in edx-platform and updated the filter definition to `CourseEnrollmentAPIRenderStarted`.
Please have a look at them and let me know of any further changes.

felipemontoya (approved):
Thanks @jajjibhai008. I would say codewise this looks good now.

I don't have access to any of the Jira tickets you linked so I have no Idea about how this fullfills or not the requirement, but I'll leave my review approval

jajjibhai008:
> 
Thanks, @felipemontoya for your kind and experienced reviews.
As I don't have access to merge it, please merge it for me, so that I can use these changes.

felipemontoya:
Could someone else from your team that has access to the jira tickets give a review as well? I have no context from the requirements other that what is in this page.

muhammad-ammar (approved):
Awesome Work!

PR #156, opened 2024-04-06
feat: Adding python3.11 support.

awais786:
Adding python3.11 support.


mariajgrimaldi:
I used a similar approach here: https://github.com/openedx/openedx-filters/pull/159, so I'll close here since it's already resolved. Thank you!

PR #158, opened 2024-04-10
feat: add ORA filter for submission view rendering intervention

BryanttV:
## Description
This PR adds a filter to modify the default template of submissions in an ORA assignment.

- New filter: `ORASubmissionViewRenderStarted`
- New exception: `RenderInvalidTemplate` 

## Supporting Information
These changes are based on those proposed in the [following ADR](https://github.com/openedx/edx-ora2/pull/2182)

PR #171, opened 2024-05-28
feat: add filter for hooking XBlock Render

nsprenkle:
**Description:** Add a filter just before rendering of an XBlock scope. [See discussion forum](https://discuss.openedx.org/t/filter-in-xblock-render-pipeline/13095/5).

**Justification:** Some of our current work requires an ability to read and modify data in the XBlock render pipeline. While there *already* exist filters in the Vertical rendering pipeline, these are limited compared to the XBlock rendering pipeline in 2 important ways.

1. The vertical render does not have access to the original request object, meaning if we are otherwise unable to read/modify data on or in response to that request.
2. The vertical render pipeline does not apply to XBlock scopes outside of vertical/unit. If, for example, we ask to render an individual block, these filters will not get activated.

**JIRA:** [AU-2039](https://2u-internal.atlassian.net/browse/AU-2039) and [2037](https://2u-internal.atlassian.net/browse/AU-2037)

**Merge deadline:** List merge deadline (if any)

**Installation instructions:** n/a

**Testing instructions:**

1. Check out related edx-platform branch: [nsprenkle/xblock-render-filter](https://github.com/openedx/edx-platform/tree/nsprenkle/xblock-render-filter)
4. Install local copy of `openedx-filters`.
5. Attempt to render an xblobk: `http://{lms}/xblock/block-v1:{block-id}?view=student_view`
6. Verify filter step is run.

**Reviewers:**
- [x] @felipemontoya 

**Merge checklist:**
- [x] All reviewers approved
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -546,6 +546,46 @@ class CourseEnrollmentQuerysetRequested(OpenEdxPublicFilter):
         return data.get("enrollments")
 
 
+class RenderXBlockStarted(OpenEdxPublicFilter):
+    """
+    Filter in between context generation and rendering of XBlock scope.
+    """
+
+    filter_type = "org.openedx.learning.xblock.render.started.v1"
+
+    class PreventXBlockBlockRender(OpenEdxFilterException):
+        """
+        Custom class used to prevent the XBlock from rendering for the user.
+        """
+
+    class RenderCustomResponse(OpenEdxFilterException):
+        """
+        Custom class used to stop the XBlock rendering process and return a custom response.
+        """
+
+        def __init__(self, message, response=None):
+            """
+            Override init that defines specific arguments used in the XBlock render process.
+
+            Arguments:
+                message: error message for the exception.
+                response: custom response which will be returned by the XBlock render view.
+            """
+            super().__init__(message, response=response)
+
+    @classmethod
+    def run_filter(cls, context, student_view_context):
+        """
+        Execute a filter with the specified signature.
+
+        Arguments:
+            context (dict): rendering context values like is_mobile_app, show_title, etc.
+            student_view_context (dict): context passed to the student_view of the block context
+        """
+        data = super().run_pipeline(context=context, student_view_context=student_view_context)
+        return data.get("context"), data.get("student_view_context")
+
+
 class VerticalBlockRenderCompleted(OpenEdxPublicFilter):
     """
     Custom class used to create filters to act on vertical block rendering completed.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,14 @@ Change Log
 Unreleased
 ----------
 
+[1.9.0] - 2024-06-14
+--------------------
+
+Added
+~~~~~~~
+
+* RenderXBlockStarted filter added which allows reading / modifying of XBlock context prior to render.
+
 [1.8.1] - 2024-04-12
 --------------------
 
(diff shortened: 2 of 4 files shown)


felipemontoya (commented):
The filter definition seems reasonable to me. Given the location you are planning to use the filter, it looks to me more similar to a `RenderXBlockCompleted`.
I left this comment at the commit you shared. https://github.com/openedx/edx-platform/commit/28e160108218c7f4c85f3fae84eb825d703c3a15#commitcomment-142615584

nsprenkle:
> The filter definition seems reasonable to me. Given the location you are planning to use the filter, it looks to me more similar to a `RenderXBlockCompleted`. I left this comment at the commit you shared. [openedx/edx-platform@28e1601#commitcomment-142615584](https://github.com/openedx/edx-platform/commit/28e160108218c7f4c85f3fae84eb825d703c3a15#commitcomment-142615584)



> The filter definition seems reasonable to me. Given the location you are planning to use the filter, it looks to me more similar to a `RenderXBlockCompleted`. I left this comment at the commit you shared. [openedx/edx-platform@28e1601#commitcomment-142615584](https://github.com/openedx/edx-platform/commit/28e160108218c7f4c85f3fae84eb825d703c3a15#commitcomment-142615584)

Have updated filter names per your suggestion.

nsprenkle:
@felipemontoya / @ormsbee , looking for re-review when you have time :)

openedx_filters/learning/filters.py
```diff
@@ -546,6 +546,31 @@ def run_filter(cls, enrollments):
...
 
 
+class RenderXBlockStarted(OpenEdxPublicFilter):
+    """
+    Filter in between context generation and rendering of XBlock scope.
+    """
+
+    filter_type = "org.openedx.learning.xblock.render.started.v1"
+
+    class PreventXBlockBlockRender(OpenEdxFilterException):
```
felipemontoya, line 556: In a comment at the branch in edx-platform I commented about it being common behavior that we allow filter users to modify what is shown to students. https://github.com/openedx/edx-platform/commit/28e160108218c7f4c85f3fae84eb825d703c3a15#r142727965

Do you have a reason for not doing that? If you agree that this should be done I would suggest we create a second exception here similar to the AccountSettingRenderStarted filter does. https://github.com/openedx/openedx-filters/blob/0738e99c60cba48e36df3731200f9b34be04439c/openedx_filters/learning/filters.py#L53C1-L67C57

felipemontoya (changes requested):
The filter looks good, but in order to give users of this filter more range of action I suggest we allow them to render something controllable and not only a 500 error.

openedx_filters/learning/filters.py
nsprenkle: Ah, okay, I didn't realize that comment also applied here in the creation of a new exception. I'll go ahead and add that.

nsprenkle: Have updated here and in edx-platform copy https://github.com/openedx/edx-platform/pull/34865/commits/6ba29598fc6ace373303ef7c889c062c85fac1ef

nsprenkle:
@felipemontoya , ready for re-review

felipemontoya (approved):
Brilliant. Thanks a lot @nsprenkle. I think the filter is ready.

@mariajgrimaldi correct me please if I'm wrong. To merge we need to do a minor bump and add the changelog, right?

such as with https://github.com/openedx/openedx-filters/pull/158/files

mariajgrimaldi:
@nsprenkle @felipemontoya: yes, that's all we need! Thank you both.

felipemontoya (approved):

PR #200, opened 2024-07-18
feat: adds CourseAboutPageURLRequested and LMSPageURLRequested filters

jignaciopm:
**Description:** Describe in a couple of sentences what this PR adds

This adds the TenantAwareLinkRenderStarted filter which receive a string context with LMS url to allowing the filter pipeline to return the information that we need to filter, in this case a tenant aware link from Studio to LMS.

An example of use is in [eox-tenant](https://github.com/eduNEXT/eox-tenant) custom filter step [pipeline](https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/filters/pipeline.py), where it takes the LMS_ROOT_URL and return only the tenant aware link in the respective site.

**JIRA:** [DS-1003](https://edunext.atlassian.net/browse/DS-1003)

**Dependencies:** 

https://github.com/openedx/edx-platform/pull/35142

**Merge deadline:** N/A

**Installation instructions:** 

Be sure to use edx-platform branch with changes https://github.com/openedx/edx-platform/pull/35142

**Testing instructions:**

The change can be tested with testing instructions in https://github.com/eduNEXT/eox-tenant/pull/156.

**Reviewers:**
- [ ] @mariajgrimaldi 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** 
N/A

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,15 @@ Unreleased
 ----------
 * Configuration for automatic filters docs generation.
 
+[1.11.0] - 2024-09-30
+--------------------
+
+Added
+~~~~~
+
+* CourseAboutPageURLRequested filter added which can be used to modify the URL for the course about organization.
+* LMSPageURLRequested filter added which can be used to modify the URL for the LMS page organization.
+
 [1.10.0] - 2024-09-20
 --------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,16 +16,16 @@ Unreleased
 * Configuration for automatic filters docs generation.
 
 [1.11.0] - 2024-09-30
---------------------
+---------------------
 
 Added
 ~~~~~
 
-* CourseAboutPageURLRequested filter added which can be used to modify the URL for the course about organization.
-* LMSPageURLRequested filter added which can be used to modify the URL for the LMS page organization.
+* CourseAboutPageURLRequested filter added which can be used to modify the URL for the course about page.
+* LMSPageURLRequested filter added which can be used to modify the URL for the LMS page.
 
 [1.10.0] - 2024-09-20
---------------------
+---------------------
 
 Added
 ~~~~~
--- a/openedx_filters/__init__.py
+++ b/openedx_filters/__init__.py
@@ -3,4 +3,4 @@ Filters of the Open edX platform.
 """
 from openedx_filters.filters import *
 
-__version__ = "1.10.0"
+__version__ = "1.11.0"
--- a/openedx_filters/course_authoring/__init__.py
+++ b/openedx_filters/course_authoring/__init__.py
@@ -1,0 +1,6 @@
+"""
+Package where events related to the ``content_authoring`` subdomain are implemented.
+
+The ``content_authoring`` subdomain corresponds to {Architecture Subdomain} defined in OEP-41.
+https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0041-arch-async-server-event-messaging.html
+"""
--- a/openedx_filters/course_authoring/filters.py
+++ b/openedx_filters/course_authoring/filters.py
@@ -1,0 +1,25 @@
+"""
+Package where filters related to the Course Authoring architectural subdomain are implemented.
+"""
+
+from openedx_filters.tooling import OpenEdxPublicFilter
+
+
+class LMSPageURLRequested(OpenEdxPublicFilter):
+    """
+    Custom class used to get lms page url filters and its custom methods.
+    """
+
+    filter_type = "org.openedx.course_authoring.lms.page.url.requested.v1"
+
+    @classmethod
+    def run_filter(cls, url, org):
+        """
+        Execute a filter with the signature specified.
+
+        Arguments:
+            url (str): the URL of the page to be modified.
+            org (str): Course org filter used as context data to get LMS configurations.
+        """
+        data = super().run_pipeline(url=url, org=org)
+        return data.get("url"), data.get("org")
--- a/openedx_filters/course_authoring/tests/__init__.py
+++ b/openedx_filters/course_authoring/tests/__init__.py
@@ -1,0 +1,3 @@
+"""
+Package where unittest for authoring subdomain filters are located.
+"""
--- a/openedx_filters/course_authoring/tests/test_filters.py
+++ b/openedx_filters/course_authoring/tests/test_filters.py
@@ -1,0 +1,32 @@
+"""
+Tests for authoring subdomain filters.
+"""
+from unittest.mock import Mock
+
+from django.test import TestCase
+
+from openedx_filters.course_authoring.filters import LMSPageURLRequested
+
+
+class TestCourseAuthoringFilters(TestCase):
+    """
+    Test class to verify standard behavior of the filters located in rendering views.
+    You'll find test suites for:
+
+    - LMSPageURLRequested
+    """
+
+    def test_lms_page_url_requested(self):
+        """
+        Test LMSPageURLRequested filter behavior under normal conditions.
+
+        Expected behavior:
+            - The filter should return lms page url requested.
+        """
+        url = Mock()
+        org = Mock()
+
+        url_result, org_result = LMSPageURLRequested.run_filter(url, org)
+
+        self.assertEqual(url, url_result)
+        self.assertEqual(org, org_result)
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -798,3 +798,23 @@ class IDVPageURLRequested(OpenEdxPublicFilter):
         """
         data = super().run_pipeline(url=url)
         return data.get("url")
+
+
+class CourseAboutPageURLRequested(OpenEdxPublicFilter):
+    """
+    Custom class used to get course about page url filters and its custom methods.
+    """
+
+    filter_type = "org.openedx.learning.course_about.page.url.requested.v1"
+
+    @classmethod
+    def run_filter(cls, url, org):
+        """
+        Execute a filter with the specified signature.
+
+        Arguments:
+            url (str): the URL of the page to be modified.
+            org (str): Course org filter used as context data to get LMS configurations.
+        """
+        data = super().run_pipeline(url=url, org=org)
+        return data.get("url"), data.get("org")
--- a/openedx_filters/learning/tests/test_filters.py
+++ b/openedx_filters/learning/tests/test_filters.py
@@ -12,6 +12,7 @@ from openedx_filters.learning.filters import (
     CertificateRenderStarted,
     CohortAssignmentRequested,
     CohortChangeRequested,
+    CourseAboutPageURLRequested,
     CourseAboutRenderStarted,
     CourseEnrollmentAPIRenderStarted,
     CourseEnrollmentQuerysetRequested,
@@ -752,3 +753,26 @@ class TestIDVFilters(TestCase):
         result = IDVPageURLRequested.run_filter(url)
 
         self.assertEqual(url, result)
+
+
+class TestCourseAboutPageURLRequested(TestCase):
+    """
+    Test class to verify standard behavior of the ID verification filters.
+    You'll find test suites for:
+
+    - CourseAboutPageURLRequested
+    """
+
+    def test_lms_page_url_requested(self):
+        """
+        Test CourseAboutPageURLRequested filter behavior under normal conditions.
+        Expected behavior:
+            - The filter should return lms page url requested.
+        """
+        url = Mock()
+        org = Mock()
+
+        url_result, org_result = CourseAboutPageURLRequested.run_filter(url, org)
+
+        self.assertEqual(url, url_result)
+        self.assertEqual(org, org_result)


openedx-webhooks:
Thanks for the pull request, @jignaciopm!

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.

#### :radio_button: Provide context

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads





#### :radio_button: Get a green build

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.



#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@openedx/hooks-extension-framework`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)

### When can I expect my changes to be merged?

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*

mphilbrick211:
Hi @jignaciopm! Just flagging there's a failing check here.

openedx_filters/content_authoring/filters.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Custom class used to create lms url creation link render filters and its custom methods.
+    """
+
+    filter_type = "org.openedx.course_authoring.lms.url.creation.started.v1"
+
+    class PreventLmsUrlCreationRender(OpenEdxFilterException):
+        """
+        Custom class used to stop lms url creation link render process.
+        """
```
mariajgrimaldi, line 19: Do you imagine a case where this URL creation needs to be stopped? If not, should we consider dropping this exception? 

I understand we often add these exceptions for parity with the other implementations. However, we should implement them only when we have a use case so we don't maintain unused code. For instance, there's a similar use case for URL runtime replacement here: https://github.com/openedx/openedx-filters/commit/809bc38dc6b64be311d2c016644187fb3de961d6#diff-0c79edd111d62405dc0829ac0b0b971c73b410de87c0133a7190b3de08b40ed9R784-R800, where we didn't need any exception since the goal of the filter is to make the replacement when a condition is met. Should we consider doing the same here?

Let me know what you think!

openedx_filters/content_authoring/filters.py
```diff
@@ -0,0 +1,33 @@
...
+        Execute a filter with the signature specified.
+
+        Arguments:
+        context (str): rendering context value
+        org (str): Course org filter, this value will be used to filter out the correct lms url configuration.
+        val_name (str): Name of the key for which to return configuration value.
+        default: default value to return if key is not present in the configuration
+        """
+        data = super().run_pipeline(context=context, org=org, val_name=val_name, default=default)
+        return data.get("context")
```
mariajgrimaldi, line 33: According to the documentation and definitions we set for the framework, each filter must return the same arguments it received: https://docs.openedx.org/projects/openedx-filters/en/latest/decisions/0003-hooks-filter-tooling-pipeline.html#consequences

openedx_filters/content_authoring/tests/test_filters.py
```diff
@@ -0,0 +1,64 @@
...
+from unittest.mock import Mock
+
+from ddt import data, ddt, unpack
+from django.test import TestCase
+
+from openedx_filters.content_authoring.filters import LmsUrlCreationStarted
+
+
+@ddt
+class TestRenderingFilters(TestCase):
```
mariajgrimaldi, line 13: This filter is not located in rendering views, so is this the correct name for the test suite?

openedx_filters/content_authoring/tests/test_filters.py
```diff
@@ -0,0 +1,64 @@
...
+        Expected behavior:
+            - The filter should return lms url creation context.
+        """
+        context = Mock()
+        org = Mock()
+        val_name = Mock()
+        default = Mock()
+
+        result = LmsUrlCreationStarted.run_filter(context, org, val_name, default)
+        print(result)
```
mariajgrimaldi, line 43: Don't forget to remove this print.

mariajgrimaldi (changes requested):

openedx_filters/content_authoring/filters.py
mariajgrimaldi: Also, how are `val_name` and `default` related to the filter itself? How are they relevant?

openedx_filters/content_authoring/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Package where filters related to the content_authoring subdomain are implemented.
+
+The content_authoring subdomain corresponds to {Architecture Subdomain} defined in
+the OEP-41.
+"""
```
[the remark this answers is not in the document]
felipemontoya: This filter is difficult to place. I understand the point about it not being in the authoring context but I don't quite like the idea of having a "common" domain. This could ultimately become one of those "utils" domains that nobody really understands.

Given how we don't have a fixed lists of domains, I would place this into the "routing" domain. Signailing that this filter (and probably many others) is used to direct users from one place to another.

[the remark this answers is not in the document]
mariajgrimaldi: We do have a list of fixed subdomains: https://openedx.atlassian.net/wiki/spaces/AC/pages/663224968/edX+DDD+Bounded+Contexts

I didn't mean `common` as a new domain cause it shouldn't be, but we need a place to get definitions from. Another solution would be to place it in both subdomains, learning and authoring.

openedx_filters/content_authoring/filters.py
jignaciopm: @mariajgrimaldi Honestly no. When I was working on this task, I used old practices to parity with the other implementations and I think that is not necessary now. We can remove it.

openedx_filters/content_authoring/tests/test_filters.py
jignaciopm: @mariajgrimaldi right, maybe the correct name is `TestContentAuthoringFilters` :thinking: 
What do you think?

openedx_filters/content_authoring/filters.py
jignaciopm: Mmm @mariajgrimaldi I understand your point. That suggests something like this:

```python
def run_filter(cls, context, org):
        """
        Execute a filter with the signature specified.
        Arguments:
        context (str): rendering context value
        org (str): Course org filter, this value will be used to filter out the correct lms url configuration.
        """
        data = super().run_pipeline(
                context=context,
                org=org,
                val_name='LMS_ROOT_URL',
                default=settings.LMS_ROOT_URL
        )
        return data.get("context")
```

Or failing that apply this in https://github.com/eduNEXT/eox-tenant/pull/179/files#diff-575cd086f0a7c1d94d2fd3889136664ba0d48d24f2ab6208eaf0c693f7b6fd1cR82

openedx_filters/content_authoring/filters.py
mariajgrimaldi: Great! Thank you

openedx_filters/content_authoring/filters.py
mariajgrimaldi: I'd imagine something like:
```
def run_filter(cls, context, org):
        """
        Execute a filter with the signature specified.
        Arguments:
        context (str): rendering context value
        org (str): Course org filter, this value will be used to filter out the correct lms url configuration.
        """
        data = super().run_pipeline(context=context, org=org)
        return data.get("context"), data.get("org")
```
And in the eox-tenant's implementation, something more like:

```lms_root = configuration_helpers.get_value_for_org(org, 'LMS_ROOT_URL', settings.LMS_ROOT_URL)
```

Since this is your pipeline implementation, you could change it to fit your specific needs. Also, I think the `lms_root` argument [here](https://github.com/openedx/edx-platform/pull/35142/files#diff-b0db2096906850f5aa17a916c950277aa7a71350c123a8a7589e307043d1cf44R724) shouldn't be called `context`, which doesn't give much away. Can we use a more meaningful name?

openedx_filters/content_authoring/tests/test_filters.py
mariajgrimaldi: That's better! Thank you

openedx_filters/content_authoring/__init__.py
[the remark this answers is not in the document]
felipemontoya: What you are trying to avoid is having a filter that was defined in the the authoring domain being used in the learning domain. Is this correct?

I don't think the feeling that this filter is generic is sufficient to define it in two different domains. We only need to solve the cases we already have.
Now, if the problem arises from actually using the filter in both locations the only thing I could propose is that we define it where it is and also make a second filter in the learning domain that inherits from this.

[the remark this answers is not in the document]
mariajgrimaldi: > What you are trying to avoid is having a filter that was defined in the the authoring domain being used in the learning domain. Is this correct?

Yes.

> I don't think the feeling that this filter is generic is sufficient to define it in two different domains. We only need to solve the cases we already have.

This filter is currently being used in two subdomains, learning and authoring. That's where the problem is and what I'm trying to solve. It's not only about being generic; it's about being so generic that it can be cross-domain.

> Now, if the problem arises from actually using the filter in both locations the only thing I could propose is that we define it where it is and also make a second filter in the learning domain that inherits from this.

That's similar to what I proposed, so I guess we agree.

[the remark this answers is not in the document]
mariajgrimaldi: @jignaciopm: the more convenient approach was implementing a filter in each subdomain. When doing so, please consider:

1. The name of the filter must reflect its context. Here's a good example you could use: [IDVPageURLRequested](https://github.com/openedx/openedx-filters/commit/809bc38dc6b64be311d2c016644187fb3de961d6#diff-0c79edd111d62405dc0829ac0b0b971c73b410de87c0133a7190b3de08b40ed9R784)
2. The filter ID must contain its corresponding subdomain, learning or authoring. Please, check the wiki page to confirm which one to use: https://openedx.atlassian.net/wiki/spaces/AC/pages/663224968/edX+DDD+Bounded+Contexts
3. Each filter must have its own test suite.

Here's how the filters would look like considering what I mentioned:

```
class LMSPageURLRequested(OpenEdxPublicFilter):

    filter_type = "org.openedx.authoring.lms.page.url.requested.v1"

    @classmethod
    def run_filter(cls, target_url, org):
        data = super().run_pipeline(target_url, org=org)
        return data.get("target_url"), data.get("org")


class CourseAboutPageURLRequested(OpenEdxPublicFilter):

    filter_type = "org.openedx.authoring.course_about.page.url.requested.v1"

    @classmethod
    def run_filter(cls, target_url, org):
        data = super().run_pipeline(target_url, org=org)
        return data.get("target_url"), data.get("org")
```

mariajgrimaldi:
@jignaciopm: Thanks again for addressing our comments! I reviewed this one last time and made some changes for you to review: https://github.com/openedx/openedx-filters/commit/4b77aaefe66fd18a529e29af783fd3f366bf757a. I'll approve this PR once you apply them.

The changes are mainly focused on using `course_authoring` instead of `authoring`, so we maintain parity with openedx-events and changing `target_url` to `url` to maintain parity with [previous implementations](https://github.com/openedx/openedx-filters/commit/809bc38dc6b64be311d2c016644187fb3de961d6). I think both were suggestions from my last review. Still, after reviewing them again, these changes seemed appropriate. The quality fix is also included there :)

I'm sorry for the back and forth.

jignaciopm:
Thank you @mariajgrimaldi the change was applied there. Tell me if you agree how it looks now.

mariajgrimaldi (approved):
LGTM! Thank you again for addressing all of our comments. Please, feel free to apply the changes in the places needed. 

I'll merge this once you add an entry to the changelog and update the `openedx_filters/__init__.py` file with the new version.

PR #333, opened 2026-03-04
feat: add GradeEventContextRequested filter for grade analytics enrichment

pwnage101:
Introduces a new GradeEventContextRequested filter (org.openedx.learning.grade.context.requested.v1) that allows pipeline steps to enrich the context dict emitted with grade analytics events. Adds unit tests confirming correct filter type and pipeline behavior.

ENT-11563

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

Blocks:
* https://github.com/openedx/edx-enterprise/pull/2543
* https://github.com/openedx/openedx-platform/pull/38097


openedx_filters/learning/filters.py
```diff
@@ -1485,3 +1485,42 @@ def run_filter(cls, readonly_fields: set, user: Any) -> tuple[Any, Any]:
...
+    Trigger:
+        - Repository: openedx/openedx-platform
+        - Path: lms/djangoapps/grades/events.py
+        - Function or Method: course_grade_passed_first_time
+    """
+
+    filter_type = "org.openedx.learning.grade.context.requested.v1"
+
+    @classmethod
+    def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict | None, int | None, str | None]:
```
pwnage101: the return type should be more strict by not letting non-None input values be converted to None.
```suggestion
    def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict, int, str]:
```

openedx_filters/learning/filters.py
```diff
@@ -1485,3 +1485,42 @@ def run_filter(cls, readonly_fields: set, user: Any) -> tuple[Any, Any]:
...
+        """
+        Process the context dict using the configured pipeline steps.
+
+        Arguments:
+            context (dict): The grade event context to be enriched.
+            user_id (int): The ID of the user associated with the grade event.
+            course_id (str): The identifier of the course.
+
+        Returns:
+            tuple[dict | None, int | None, str | None]:
```
```suggestion
            tuple[dict, int, str]:
```

openedx_filters/learning/filters.py
pwnage101: While we're in here, I always thought this was a bug but was too late to comment on it.  Unless there's an exceptional reason, I think Input types should match output types.

```python
    def run_filter(cls, readonly_fields: set, user: Any) -> tuple[set, Any]:
```

openedx_filters/learning/filters.py
```diff
@@ -1485,3 +1485,42 @@ def run_filter(cls, readonly_fields: set, user: Any) -> tuple[Any, Any]:
...
+            course_id (str): The identifier of the course.
+
+        Returns:
+            tuple[dict, int, str]:
+                dict: The enriched grade event context.
+                int: The user ID.
+                str: The course identifier.
+        """
+        data = super().run_pipeline(context=context, user_id=user_id, course_id=course_id)
+        return data.get("context"), data.get("user_id"), data.get("course_id")
```
pwnage101: As discussed, lets just make this strict:
```suggestion
        return data["context"], data["user_id"], data["course_id"]
```

That way, if the pipeline calling code can't tolerate None values from a buggy filter, then at least we catch the error sooner. Now the error (KeyError) happens closer to the root issue inside the pipeline itself, rather than after the entire pipeline has finished and returned to openedx-platform, potentially throwing a TypeError/AttributeError resulting from working with None values unexpectedly.

pwnage101:
✅ LGTM

setup.py
```diff
@@ -147,7 +147,7 @@ def is_requirement(line):
     ],
     include_package_data=True,
     install_requires=load_requirements("requirements/base.in"),
-    python_requires=">=3.12",
```
kiram15, line 150: We aren't using any python 3.12 specific functionality, and this allows edx-platform to be able to pick up the new openedx-filters versions that are right now incompatible with edx-platform's python version of 3.11.

openedx_filters/learning/filters.py
```diff
@@ -1468,7 +1468,7 @@ class AccountSettingsReadOnlyFieldsRequested(OpenEdxPublicFilter):
     filter_type = "org.openedx.learning.account.settings.read_only_fields.requested.v1"
 
     @classmethod
-    def run_filter(cls, readonly_fields: set, user: Any) -> tuple[Any, Any]:
+    def run_filter(cls, readonly_fields: set, user: Any) -> tuple[set, Any]:
```
felipemontoya, line 1471: [info] this is coming from the last PR that added this filter where we agreed to this after the PR was merged.

felipemontoya (approved):
This looks very clean and the accompanying platform PR is also clear and with good structure.
I think removing one more stitch in enterprise dependency while at the same time allowing others to enrich the grade analytics is completely welcomed.

PR #335, opened 2026-03-04
feat: add DiscountEligibilityCheckRequested filter

pwnage101:
Adds a new DiscountEligibilityCheckRequested filter to the learning subdomain, enabling plugins to mark users as ineligible for LMS-controlled course discounts without coupling the core platform to enterprise-specific logic.

ENT-11564

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

Blocks:
* https://github.com/openedx/edx-enterprise/pull/2545
* https://github.com/openedx/openedx-platform/pull/38099


openedx_filters/learning/filters.py
```diff
@@ -1445,3 +1445,48 @@ def run_filter(cls, schedules: QuerySet) -> QuerySet | None:
...
+
+    filter_type = "org.openedx.learning.discount.eligibility.check.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        user: Any,
+        course_key: Any,
+        is_eligible: bool,
+    ) -> tuple:
```
pwnage101: make type more explicit.

openedx_filters/learning/filters.py
```diff
@@ -1445,3 +1445,48 @@ def run_filter(cls, schedules: QuerySet) -> QuerySet | None:
...
+    Purpose:
+        This filter is triggered during discount applicability checks, just before the
+        final eligibility decision is returned to the caller. Pipeline steps may set
+        ``is_eligible`` to ``False`` to exclude a user from receiving a discount.
+
+    Filter Type:
+        org.openedx.learning.discount.eligibility.check.requested.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
```
pwnage101: Should reference `openedx/openedx-platform`, not the legacy `openedx/edx-platform` path.

openedx_filters/learning/filters.py
```diff
@@ -1445,3 +1445,48 @@ def run_filter(cls, schedules: QuerySet) -> QuerySet | None:
...
+            is_eligible (bool): the current eligibility status before plugin evaluation.
+
+        Returns:
+            tuple[User, CourseKey, bool]:
+                - User: the Django User object (unchanged).
+                - CourseKey: the course key (unchanged).
+                - bool: the (possibly overridden) eligibility flag.
+        """
+        data = super().run_pipeline(user=user, course_key=course_key, is_eligible=is_eligible)
+        return data.get("user"), data.get("course_key"), data.get("is_eligible")
```
pwnage101: Use strict dict access here: `data["user"], data["course_key"], data["is_eligible"]`. Per the convention established on #333, `data.get(...)` lets a buggy pipeline step silently drop a key — `None` would then propagate into the caller and falsy-evaluate as ineligible. We want a `KeyError` raised inside the pipeline instead.

openedx_filters/learning/filters.py
pwnage101: Tighten the return annotation from bare `tuple` to `tuple[Any, Any, bool]` (or, better, concrete `User` / `CourseKey` types alongside fixing the input annotations below). Per #333, leaving this as a wide `tuple` allows a buggy step to return `None` for required values without the type checker flagging it.

openedx_filters/learning/filters.py
```diff
@@ -1445,3 +1445,48 @@ def run_filter(cls, schedules: QuerySet) -> QuerySet | None:
...
+        - Function or Method: can_receive_discount, can_show_streak_discount_coupon
+    """
+
+    filter_type = "org.openedx.learning.discount.eligibility.check.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        user: Any,
+        course_key: Any,
```
pwnage101: The docstring below already names `User` and `CourseKey or course object` — lift those into the annotations (`user: User, course_key: CourseKey`) so input and output align and `Any` doesn't drift in.

openedx_filters/learning/tests/test_filters.py
```diff
@@ -801,3 +802,31 @@ def test_schedule_requested(self):
...
+            "openedx_filters.tooling.OpenEdxPublicFilter.run_pipeline",
+            return_value={"user": user, "course_key": course_key, "is_eligible": False},
+        ):
+            returned_user, returned_course_key, is_eligible = DiscountEligibilityCheckRequested.run_filter(
+                user=user, course_key=course_key, is_eligible=True
+            )
+
+        self.assertFalse(is_eligible)
+        self.assertIs(returned_user, user)
+        self.assertIs(returned_course_key, course_key)
```
pwnage101: Per the minimal-test convention from #334, the canonical test for a new filter in this repo verifies inputs pass through **unchanged when no pipeline is configured** — mirror `test_schedule_requested` directly above. The override-to-`False` case exercises pipeline-step behavior and belongs alongside the pipeline-step implementation in edx-enterprise, not here.

pwnage101 (commented):
Don't forget to also bump the version.

openedx_filters/learning/filters.py
marlonkeating: Done for `CourseKey`, omitting for `User` as the type is not referenced anywhere in the file and we would need to do something like [User = get_user_model()](https://github.com/openedx/openedx-platform/blob/a4cb48f6458c1ccb50ef171d78ed27c0ce63af5b/cms/djangoapps/contentstore/helpers.py#L53), which I imagine we shouldn't do just to get a Type annotation.

pwnage101: You actually can use AbstractBaseUser from django. See my draft PR for a different ticket: https://github.com/openedx/openedx-filters/pull/336/changes#diff-0c79edd111d62405dc0829ac0b0b971c73b410de87c0133a7190b3de08b40ed9

```python
from django.contrib.auth.base_user import AbstractBaseUser
[...]
    def run_filter(cls, user: AbstractBaseUser, course_key: CourseKey) -> tuple[AbstractBaseUser, CourseKey]:
```

openedx_filters/learning/filters.py
```diff
@@ -1598,3 +1598,48 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+
+    filter_type = "org.openedx.learning.discount.eligibility.check.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        user: Any,
+        course_key: CourseKey,
+        is_eligible: bool,
+    ) -> tuple[Any, CourseKey | None, bool | None]:
```
pwnage101: ❌ Simplify the output type to remove ambiguity---None values should not be allowed.
```suggestion
    def run_filter(
        cls,
        user: AbstractBaseUser,
        course_key: CourseKey,
        is_eligible: bool,
    ) -> tuple[AbstractBaseUser, CourseKey, bool]:
```

openedx_filters/learning/filters.py
```diff
@@ -1598,3 +1598,48 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+    """
+
+    filter_type = "org.openedx.learning.discount.eligibility.check.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        user: Any,
+        course_key: CourseKey,
+        is_eligible: bool,
```
pwnage101: Nit: since it only takes _one_ pipeline step to deem the entire discount ineligible, every pipeline step implementation must take care to skip itself when `is_eligible`=False to prevent clobbering a prior step which set it to False.  This pushes complexity into 3rd party code when there's really a much simpler solution:

We could make pipeline steps simply raise an `OpenEdxFilterException` exception.  There are many examples in this same file, including `StudentRegistrationRequested.PreventRegistration`:
https://github.com/edx/openedx-filters/blob/main/openedx_filters/learning/filters.py#L161

This has the side-benefit of providing a clean way to pass error messages from the caller without polluting the `run_filter()` signature. See my in-progress work for a different ticket: https://github.com/openedx/openedx-filters/pull/336/changes

Notably:
```python
class CoursewareAccessChecksRequested(OpenEdxPublicFilter):
   [...]

    class PreventCoursewareAccess(OpenEdxFilterException):
        [...]

        def __init__(
            self,
            error_code: str,
            developer_message: str,
            user_message: str,
        ) -> None:
            super().__init__()
            self.error_code = error_code
            self.developer_message = developer_message
            self.user_message = user_message
```

openedx_filters/learning/filters.py
pwnage101: See also my comment on an older thread about AbstractBaseUser: https://github.com/openedx/openedx-filters/pull/335#discussion_r3289900524

openedx_filters/learning/filters.py
```diff
@@ -1782,3 +1782,68 @@ def run_filter(cls, user: Any, course_key: CourseKey) -> tuple[Any, CourseKey]:
...
+        def __init__(
+            self,
+            message: str,
+        ) -> None:
+            """
+            Initialize with a human-readable reason why the user is ineligible for a discount.
+            """
+            if not isinstance(message, str) or not message.strip():
+                raise ValueError("message must be a non-blank string with reason for discount ineligibility")
+            super().__init__(message=message)
```
pwnage101: Since we're only collecting a message, and a `message` kwarg is [already inherited from OpenEdxFilterException](https://github.com/pwnage101/openedx-filters/blob/04b55915dc48b4d171f180324ebd32ef75cc2915/openedx_filters/exceptions.py#L26), there's no need for this entire init override.

openedx_filters/learning/filters.py
```diff
@@ -1782,3 +1782,68 @@ def run_filter(cls, user: Any, course_key: CourseKey) -> tuple[Any, CourseKey]:
...
+
+    @classmethod
+    def run_filter(
+        cls,
+        user: Any,
+        course_key: CourseKey,
+        is_eligible: bool,
```
pwnage101: is_eligible is redundant with the `DiscountIneligible` exception.
```suggestion
```

pwnage101 (commented):
LGTM, but needs a rebase since 3.6.0 was already released

PR #336, opened 2026-03-05
feat: add CoursewareViewStarted, CourseStartDateValidationFailed, and CoursewareAccessChecksRequested filters

pwnage101:
Plugins can now supply a courseware view redirect URL, inject specific error
payloads when course start-date validation fails, or deny courseware access
with a priority (non-bypassable) error.

ENT-11544

---

Blocks:
* https://github.com/openedx/edx-enterprise/pull/2548
* https://github.com/edx/edx-platform/pull/338
* https://github.com/edx/edx-platform/pull/339
* https://github.com/edx/edx-platform/pull/340
* https://github.com/openedx/openedx-platform/pull/38102

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,15 +33,26 @@ Unreleased
 ----------
 .. scriv-insert-here
 
+.. _changelog-3.5.0:
+
+[v3.5.0] - 2026-06-02
+---------------------
+
+Added
+~~~~~
+
+* Added ``CoursewareViewStarted``, ``CourseStartDateValidationFailed``, and ``CoursewareAccessChecksRequested`` filters to the learning subdomain.
+
 [3.4.1] - 2026-05-26
+--------------------
 
 Changed
-~~~~~
+~~~~~~~
 
 * Update ``InstructorDashboardTabsRequested`` to match ``run_filter`` return pattern of other filters. by @holaontiveros
 
-
 [3.4.0] - 2026-05-07
+--------------------
 
 Added
 ~~~~~
@@ -53,7 +64,7 @@ Added
 See the fragment files in the `changelog.d directory`_.
 .. _changelog.d directory: https://github.com/openedx/openedx-filters/tree/master/changelog.d
 
-[3.3.0] - 2025-04-17
+[3.3.0] - 2026-04-17
 --------------------
 
 Changed
@@ -61,7 +72,7 @@ Changed
 
 * Expand Python version compatibility
 
-[3.2.0] - 2025-04-13
+[3.2.0] - 2026-04-13
 --------------------
 
 Changed
@@ -69,7 +80,7 @@ Changed
 
 * Added GradeEventContextRequested filter
 
-[3.1.0] - 2025-04-06
+[3.1.0] - 2026-04-06
 --------------------
 
 Changed
@@ -77,13 +88,38 @@ Changed
 
 * Added AccountSettingsReadOnlyFieldsRequested filter
 
-[3.0.0] - 2025-03-19
+[3.0.0] - 2026-03-19
 --------------------
 
 Changed
 ~~~~~~~
 
 * Dropped Python 3.11 support
+
+[2.1.0] - 2025-04-28
+--------------------
+
+Added
+~~~~~
+
+* New documentation section with naming suggestions for creating filters. (by @mariajgrimaldi)
+
+* Migrate to ``scriv`` for manage changelog. (by @bryanttv)
+
+Changed
+~~~~~~~
+
+* Updated documentation titles and styles to follow the Open edX style guide. (by @Apgomeznext)
+
+* Updated the glossary section in the documentation. (by @Apgomeznext)
+
+* Improved how-to guides including updates to filters, samples, and formatting. (by @mariajgrimaldi)
+
+* Updated edX RTD link to Open edX RTD. (by @sarina)
+
+* Replaced the deprecated ``pydocstyle`` library with ``ruff``. (by @bryanttv)
+
+* Improved documentation of filter configuration including details on configuration formats. (by @bryanttv)
 
 [2.0.1] - 2025-02-18
 --------------------
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -2,6 +2,7 @@
 Package where filters related to the learning architectural subdomain are implemented.
 """
 
+from datetime import datetime
 from typing import Any, Optional
 
 from django.db.models.query import QuerySet
@@ -25,7 +26,7 @@ class AccountSettingsRenderStarted(OpenEdxPublicFilter):
         org.openedx.learning.student.settings.render.started.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: openedx/core/djangoapps/user_api/accounts/settings_views.py
         - Function or Method: account_settings
 
@@ -142,7 +143,7 @@ class StudentRegistrationRequested(OpenEdxPublicFilter, SensitiveDataManagementM
         org.openedx.learning.student.registration.requested.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: openedx/core/djangoapps/user_authn/views/register.py
         - Function or Method: RegistrationView.post
     """
@@ -195,7 +196,7 @@ class StudentLoginRequested(OpenEdxPublicFilter):
         org.openedx.learning.student.login.requested.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: openedx/core/djangoapps/user_authn/views/login.py
         - Function or Method: login_user
     """
@@ -260,7 +261,7 @@ class CourseEnrollmentStarted(OpenEdxPublicFilter):
         org.openedx.learning.course.enrollment.started.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: common/djangoapps/student/models/course_enrollment.py
         - Function or Method: enroll
     """
@@ -310,7 +311,7 @@ class CourseUnenrollmentStarted(OpenEdxPublicFilter):
         org.openedx.learning.course.unenrollment.started.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: common/djangoapps/student/models/course_enrollment.py
         - Function or Method: unenroll
     """
@@ -352,7 +353,7 @@ class CertificateCreationRequested(OpenEdxPublicFilter):
         org.openedx.learning.certificate.creation.requested.v1
 
     Trigger:
-        - Repository: openedx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: lms/djangoapps/certificates/generated_certificate.py
         - Function or Method: _generate_certificate_task
     """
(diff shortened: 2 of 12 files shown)


openedx_filters/learning/filters.py
```diff
@@ -1445,3 +1445,45 @@ def run_filter(cls, schedules: QuerySet) -> QuerySet | None:
...
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: lms/djangoapps/courseware/decorators.py
+        - Function or Method: courseware_view_redirect
+    """
+
+    filter_type = "org.openedx.learning.courseware.view.redirect_url.requested.v1"
+
+    @classmethod
+    def run_filter(cls, redirect_urls: list, request: Any, course_key: Any) -> tuple:
```
pwnage101: make return type more explicit. tuple of what?

openedx_filters/learning/filters.py
```diff
@@ -1601,3 +1603,188 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+
+    filter_type = "org.openedx.learning.courseware.view.started.v1"
+
+    class RedirectToUrl(OpenEdxFilterException):
+        """Raised by a pipeline step to redirect the user before the courseware view renders."""
+
+        def __init__(self, url: str) -> None:
+            """Store the redirect URL on the exception instance."""
+            super().__init__()
+            self.url = url
```
kiram15: This differs from other exceptions in the codebase that typically pass a message, can we mimic this behavior?
https://github.com/pwnage101/openedx-filters/blob/b80e6cedade87031fcd9553d1fd102cceea3564d/openedx_filters/learning/filters.py#L60

CHANGELOG.rst
```diff
@@ -53,38 +64,63 @@ Added
...
+[3.2.0] - 2026-04-13
 --------------------
 
 Changed
 ~~~~~~~
 
 * Added GradeEventContextRequested filter
 
-[3.1.0] - 2025-04-06
+[3.1.0] - 2026-04-06
```
kiram15, line 83: Well this is embarrassing

openedx_filters/learning/filters.py
```diff
@@ -1601,3 +1603,188 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+        ``StartDateFiltersError`` from the exception's fields.
+
+        The first pipeline step to raise wins; later steps do not run.
+
+    Filter Type:
+        org.openedx.learning.course.start_date.validation_failed.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: lms/djangoapps/courseware/access_utils.py
```
kiram15, line 1675: Path is incorrect, should be [enterprise/filters/courseware.py‎](https://github.com/openedx/edx-enterprise/pull/2548/changes#diff-3a369eb3e8f39b0f443dd5dc5bd5addbb1c4f6bf83ef07bb4b3a30a882f13dc8)

pwnage101: The path you shared is an edx-enterprise path to the pipeline step, but what's actually needed is the "Trigger" a.k.a. the [call site of the filter](https://github.com/openedx/openedx-platform/blob/c0e2e423ebddf0854ec45b5e79a040688399e616/lms/djangoapps/courseware/access_utils.py#L103).

openedx_filters/learning/filters.py
pwnage101: Updated all added filter exceptions to require a `message` argument.

pwnage101:
@felipemontoya this is now ready for your review.  For reference, the proposed call sites for all 3 added filters can be found in the [draft platform PR](https://github.com/openedx/openedx-platform/pull/38102).

Also, other unrelated repo cleanup stuff:

* Updated all outdated references `openedx/edx-platform` -> `openedx/openedx-platform`
* Fixed CHANGELOG.rst:
  * Deleted all the fragments under changelog.d/ since they are supposed to be deleted automatically by `make changelog`.
  * Manually incorporated those deleted fragments into CHANGELOG.rst.
  * Fixed a bunch of broken and malformed changelog entries (missing headers, wrong dates, etc.)
  * Added my own changelog entry using the new `make changelog-entry` / `make changelog` mechanism.

marlonkeating (approved):

felipemontoya:
Reviewing now.

For future reference @pwnage101, whenever you need to do a lot of cosmetic changes its better to keep those in a separate PR. Those I can merge easily and that makes reviewing the actual change in the filters easier.

openedx_filters/learning/filters.py
```diff
@@ -1601,3 +1602,183 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
         """
         data = super().run_pipeline(context=context, user_id=user_id, course_id=course_id)
         return data["context"], data["user_id"], data["course_id"]
+
+
+class CoursewareViewStarted(OpenEdxPublicFilter):
```
felipemontoya, line 1607: This filter makes a lot of sense to me.

If I understand correctly this goes first in the stack, right?

```
1. CoursewareViewStarted
2. VerticalBlockChildRenderStarted
3. Loop over RenderXBlockStarted
4. VerticalBlockRenderCompleted

File: openedx_filters/learning/filters.py [hunk 6]
```diff
@@ -1601,3 +1602,183 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+        ``StartDateFiltersError`` from the exception's fields.
+
+        The first pipeline step to raise wins; later steps do not run.
+
+    Filter Type:
+        org.openedx.learning.course.start_date.validation_failed.v1
+
+    Trigger:
+        - Repository: openedx/openedx-platform
+        - Path: lms/djangoapps/courseware/access_utils.py
```
felipemontoya, line 1675: Do you think this must connect with the RBAC project in any meaningful way?

@mariajgrimaldi or @BryanttV do you see any cause for concern with a filter like this?

openedx_filters/learning/filters.py
```diff
@@ -1601,3 +1602,183 @@ def run_filter(cls, context: dict, user_id: int, course_id: str) -> tuple[dict,
...
+        data = super().run_pipeline(
+            course_key=course_key,
+            start_date=start_date,
+        )
+        return data["course_key"], data["start_date"]
+
+
+class CoursewareAccessChecksRequested(OpenEdxPublicFilter):
+    """
+    Filter triggered during courseware access checks so plugins can deny access.
```
felipemontoya, line 1728: This filter also makes a lot of sense to me. I'd also like to see how this interacts with RBAC for future-proofing purposes.

felipemontoya (approved):
Overall the 3 filters make are well defined, correctly documented and are new places in the platform cycle not covered by previous filters. I'd like to hear from the RBAC devs before merging but otherwise I approve.

openedx_filters/learning/filters.py
pwnage101: To be completely honest, I don't 100% know.  My goal with this ticket is a 1:1 replacement, so I'm pretty laser focused on replacing the exact call sites.  That said, filter naming is important, so it is a good idea to understand the code path to come up with a self-documenting name.

The best I can tell, there are 2 code paths that can trigger a DSC redirect when a learner views courseware (legacy vs. MFE), and they use different filters:

**Legacy (server-rendered) courseware views:**
1. CoursewareViewStarted (view decorator / WikiAccessMiddleware)
   -  If DSC redirect is needed, returns an HTTP redirect here.
3. loop over VerticalBlockChildRenderStarted
4. VerticalBlockRenderCompleted

**Learning MFE — two separate requests:**
- Request A — course_home_api/course_metadata BFF call:
  1. CoursewareAccessChecksRequested (from check_course_access)
     - If DSC redirect is needed, returns the error_code = `data_sharing_access_required` and developer_message = `<consent URL>` in API response.
     - MFE reads developerMessage as the consent redirect URL, performs redirect via javascript.
- Request B — render_xblock:
  1. RenderXBlockStarted
  2. loop over VerticalBlockChildRenderStarted
  3. VerticalBlockRenderCompleted

openedx_filters/learning/filters.py
pwnage101: I'm not sure, this is the first I've heard of the "RBAC project" ([openedx/openedx-authz](https://github.com/openedx/openedx-authz/)), having only contributed to [openedx/edx-rbac](https://github.com/openedx/edx-rbac) which we use extensively to manage role-based authorization for all of our enterprise IDAs.

pwnage101:
@felipemontoya please consider moving forward with merging this if we don't hear back from @mariajgrimaldi or @BryanttV as it is designed to be a 1-for-1 replacement of existing logic, and the filters are generally reusable beyond access control purposes.

openedx_filters/learning/tests/test_filters.py
```diff
@@ -966,3 +971,112 @@ def test_prevent_tabs_generation_exception(self, exception_class, attributes):
...
+            - The filter returns ``course_key`` unchanged when no pipeline steps raise.
+        """
+        course_key = CourseKey.from_string("course-v1:edX+DemoX+Demo_Course")
+        view_name = "test_view"
+        result_course_key, result_view_name = CoursewareViewStarted.run_filter(
+            course_key=course_key,
+            view_name=view_name,
+        )
+        assert result_course_key == course_key
+        assert result_view_name == view_name
```
BryanttV, line 995: Please use the `TestCase` methods

```suggestion
        self.assertEqual(result_course_key, course_key)
        self.assertEqual(result_view_name, view_name)
```

openedx_filters/learning/filters.py
BryanttV: Thanks for tagging me, @felipemontoya, and sorry for the delay.

Reviewing where this filter is intended to be used in openex-platform, I don't see any concerns that directly affect the new RBAC system.

The code validates whether the user has the `CourseBetaTesterRole`, but thanks to the legacy roles compatibility layer, it will continue working without any issues.

openedx_filters/learning/filters.py
BryanttV: I also don't see any concerns with applying the current filter for the current/future RBAC implementation

openedx_filters/learning/tests/test_filters.py
pwnage101: I'm happy to change this, but I want to push back on this once. Some openedx repositories have started adopting `ruff` linting, including openedx/openedx-platform, which [fails when encountering unittest-style assertions](https://github.com/astral-sh/ruff/blob/main/crates/ruff_linter/src/rules/flake8_pytest_style/rules/assertion.rs#L174).  Today, all new openedx-platform PRs and PRs across some other openedx repositories require the newer pytest-style assertions (`assert x == y`).

My goal in this PR is to start the transition to pytest-style assertions for this repo to minimize changes once it is adopted in CI.

pwnage101: Random memory: I worked in the edX office when we switched from `nose` to `pytest`.  The campaign was called "Blow Your Nose", and I remember sitting in the conference room watching Ned Batchelder's presentation about the transition, which included a slide about switching edx-platform to use pytest-style assertions.  This was probably 9 years ago!

To this day, I'm carrying forward the long tradition of blowing my nose.

openedx_filters/learning/filters.py
felipemontoya: This is what I needed to know. I'll be moving forward with the merge

openedx_filters/learning/tests/test_filters.py
BryanttV: Haha, I didn't know about that change. Thanks for mentioning it! I'll keep the tradition going from now on as well

PR #337, opened 2026-03-05
feat: add filters to support logistration integration

pwnage101:
* Added new `authentication` architecture subdomain, and ADR to explain why it is needed.
* Added six new authentication-related filters:
  * `LogistrationViewContextGenerated`
  * `AuthnMFEContextGenerated`
  * `LoginAltRedirectURLRequested`
  * `LoginFormGenerated`
  * `RegistrationFormGenerated`
  * `LogistrationViewRenderCompleted`
* Added several new structural types:
  * `FormDescriptionProtocol` - declares the minimal surface the form filters pass to pipeline steps.
  * `ProviderConfigProtocol` - declares the minimal surface of the third-party auth provider configuration that the form filters pass to pipeline steps.
  * `RunningPipeline` - authentication pipeline state.

[ENT-11568](https://2u-internal.atlassian.net/browse/ENT-11568)

---

Integration Testing
---

See https://github.com/openedx/edx-enterprise/pull/2551#issue-4024890666

---

Related:
* https://github.com/openedx/edx-enterprise/pull/2662
* https://github.com/openedx/openedx-filters/pull/337
* https://github.com/openedx/edx-enterprise/pull/2551
* https://github.com/edx/edx-platform/pull/407
* https://github.com/openedx/openedx-platform/pull/38105
* https://github.com/edx/devstack/pull/227

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,32 @@ Unreleased
 ----------
 .. scriv-insert-here
 
-.. _changelog-3.7.0:
+.. _changelog-3.9.0:
+
+[3.9.0] - 2026-08-04
+--------------------
+
+Added
+~~~~~
+
+* Added new ``authentication`` architecture subdomain
+
+* Added six new authentication-related filters:
+
+  * ``LogistrationViewContextGenerated``
+  * ``AuthnMFEContextGenerated``
+  * ``LoginAltRedirectURLRequested``
+  * ``LoginFormGenerated``
+  * ``RegistrationFormGenerated``
+  * ``LogistrationViewRenderCompleted``
+
+* Added three new structural types used by several of the above filters:
+
+  * ``FormDescriptionProtocol`` - declares the minimal surface of the form
+    description that the form filters pass to pipeline steps.
+  * ``ProviderConfigProtocol`` - declares the minimal surface of the third-party
+    auth provider configuration that the form filters pass to pipeline steps.
+  * ``RunningPipeline`` - authentication pipeline state.
 
 [3.8.0] - 2026-07-07
 ---------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,12 +37,18 @@ extensions = [
     'sphinx_copybutton',
     'sphinx.ext.graphviz',
     'sphinxcontrib.mermaid',
+    'myst_parser',
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
     'sphinx.ext.intersphinx',
     'sphinx.ext.napoleon',
     'sphinx.ext.linkcode',
 ]
+
+# Render fenced ```mermaid code blocks in Markdown (MyST) sources through the
+# sphinxcontrib.mermaid directive, so the same fences render both on GitHub and
+# in the Sphinx-built docs.
+myst_fence_as_directive = ["mermaid"]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/decisions/0008-authentication-subdomain.md
+++ b/docs/decisions/0008-authentication-subdomain.md
@@ -1,0 +1,100 @@
+# 8. The authentication architecture subdomain
+
+## Status
+
+Accepted
+
+## Context
+
+Filters are grouped by architecture subdomain (see
+[ADR-4](0004-filters-naming-and-versioning.rst)).
+Until now, the filters covering how users sign in and register —
+`StudentLoginRequested` and `StudentRegistrationRequested` — have lived in the
+`learning` subdomain, because that was the only user-facing subdomain available
+at the time.
+
+A new set of filters covering the login and registration *user experience* is
+being introduced: hooks around the login/registration page context and render
+lifecycle, the login and registration form descriptions, the authentication MFE
+(`frontend-app-authn`) context, and the post-login redirect. These needed a
+home, and `learning` is a poor fit: authentication gates access for *every* user
+of the platform — learners and content authors alike — so it is a distinct
+bounded context rather than a learning activity.
+
+## Decision
+
+We will introduce a new "Authentication" architecture subdomain, implemented in
+`openedx_filters/authentication/` with filter types under
+`org.openedx.authentication.*`. From now onward, new authentication-related
+filters should be added here.
+
+Docs: This new subdomain will be added to the [Architecture Subdomains Reference](../reference/architecture-subdomains.rst)
+alongside "Learning" and "Content Authoring".
+
+The following filters are added to it:
+
+- `LogistrationViewContextGenerated` — enrich the legacy (server-rendered) login/registration page context.
+- `LogistrationViewRenderCompleted` — modify the rendered legacy login/registration page response.
+- `AuthnMFEContextGenerated` — enrich the authentication MFE context.
+- `LoginFormGenerated` — augment the generated login form description.
+- `RegistrationFormGenerated` — augment the generated registration form description.
+- `LoginAltRedirectURLRequested` — choose an alternative post-login redirect.
+
+The diagram below shows when each authn-related filter fires within the authn flow.
+
+```mermaid
+flowchart TD
+    classDef filter fill:#e8f0fe,stroke:#1a73e8,color:#0a3069
+
+    ROUTES["/login and /register routes"]
+    LEGACY["Legacy logistration view<br/>(server-rendered)"]
+    MFE["Authn MFE<br/>(frontend-app-authn)"]
+
+    LogistrationViewContextGenerated["<b>LogistrationViewContextGenerated</b><br/><i>enrich legacy page context</i>"]:::filter
+    LogistrationViewRenderCompleted["<b>LogistrationViewRenderCompleted</b><br/><i>post-render response hook,<br/>useful for setting extra cookies</i>"]:::filter
+    AuthnMFEContextGenerated["<b>AuthnMFEContextGenerated</b><br/><i>enrich MFE context</i>"]:::filter
+    LoginFormGenerated["<b>LoginFormGenerated</b><br/><i>augment login form fields</i>"]:::filter
+    RegistrationFormGenerated["<b>RegistrationFormGenerated</b><br/><i>augment registration form fields</i>"]:::filter
+    StudentRegistrationRequested["<b>StudentRegistrationRequested</b><br/><i>hook for blocking registration</i>"]:::filter
+    StudentLoginRequested["<b>StudentLoginRequested</b><br/><i>hook for blocking login</i>"]:::filter
+    LoginAltRedirectURLRequested["<b>LoginAltRedirectURLRequested</b><br/><i>force alternative post-login redirect</i>"]:::filter
+
+    RENDERED["Legacy page rendered"]
+    FORMS["Login/registration<br/>FormDescription generation"]
+    LOGINPOST["Login POST endpoint"]
+    REGPOST["Registration POST endpoint"]
+    DEST["Post-auth destination,<br/><i>URL possibly overridden by filter</i>"]
+
+    ROUTES -- "AuthN MFE disabled for flow" --> LEGACY
+    ROUTES -- "AuthN MFE enabled for flow" --> MFE
+    LEGACY --> LogistrationViewContextGenerated --> RENDERED --> LogistrationViewRenderCompleted
+    LogistrationViewRenderCompleted -- "Get FormDescription via python API" --> FORMS
+    MFE --> AuthnMFEContextGenerated
+    AuthnMFEContextGenerated -- "Get FormDescription via REST API" --> FORMS
+    %% invisible edge: pin AuthnMFEContextGenerated to the same rank as
+    %% LogistrationViewContextGenerated so both *ContextGenerated filters render on the same level
+    AuthnMFEContextGenerated ~~~ RENDERED
+    FORMS -- "is /login route" --> LoginFormGenerated
+    FORMS -- "is /register route" --> RegistrationFormGenerated
+    %% invisible edge: pin LoginFormGenerated to the same rank as
+    %% RegistrationFormGenerated so both *FormGenerated filters render on the same level
+    LoginFormGenerated ~~~ REGPOST
+    LoginFormGenerated -- "submit login" --> LOGINPOST
+    RegistrationFormGenerated -- "submit registration" --> REGPOST
+    REGPOST --> StudentRegistrationRequested
+    LOGINPOST --> StudentLoginRequested
+    StudentLoginRequested -- "AuthN MFE enabled and the request is for first-party auth" --> LoginAltRedirectURLRequested
+    StudentLoginRequested -- "otherwise" --> DEST
+    LoginAltRedirectURLRequested --> DEST
+    StudentRegistrationRequested -- "account created and logged in" --> DEST
+```
+
+## Consequences
+
+- Future authentication-related filters have a clear, accurately named home and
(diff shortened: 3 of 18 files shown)


openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,293 @@ def run_filter(
...
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: openedx/core/djangoapps/user_authn/views/login_form.py
+        - Function or Method: login_and_registration_form
+    """
+
+    filter_type = "org.openedx.learning.logistration.context.requested.v1"
+
+    @classmethod
+    def run_filter(cls, context: dict) -> dict | None:
```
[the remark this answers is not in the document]
pwnage101: I'll fix.

CHANGELOG.rst
```diff
@@ -33,6 +33,23 @@ Unreleased
...
+* Added new ``LogistrationContextRequested`` filter
+* Added new ``PostLoginRedirectURLRequested`` filter
+* Added new ``LoginFormTPAOverridesRequested`` filter
+* Added new ``RegistrationFormTPAOverridesRequested`` filter
+* Added ``FormDescriptionProtocol`` and ``ProviderConfigProtocol`` structural types
+  declaring the minimal surface the form-override filters pass to pipeline steps
+* Added new ``LogistrationResponseRendered`` filter
+* Added new ``LogistrationMFERedirectRequested`` filter
+
 .. _changelog-3.7.0:
```
kiram15: Not directly part of your PR, but shouldn't this be 3.8.0?

openedx_filters/learning/tests/test_filters.py
```diff
@@ -1168,3 +1174,151 @@ def test_course_mode_price_requested(self):
...
         assert price == result_price
+
+
+class TestLogistrationContextRequestedFilter(TestCase):
+    """
+    Tests for the LogistrationContextRequested filter.
+    """
+
+    def test_filter_type(self):
+        self.assertEqual(
```
kiram15: We should be using python asserts instead of unittest ones

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+        Returns:
+            tuple[FormDescriptionProtocol, dict, ProviderConfigProtocol | None]: the (possibly
+            modified) form description, the running pipeline, and the current provider.
+        """
+        data = super().run_pipeline(
+            form_desc=form_desc,
+            running_pipeline=running_pipeline,
+            current_provider=current_provider,
+        )
+        return data["form_desc"], data["running_pipeline"], data["current_provider"]
```
iloveagent57: is this identical to the `LoginFormTPAOverridesRequested.run_filter()` definition?

iloveagent57 (approved):
Looks good, one small suggestion to DRY things up a tiny bit.

openedx_filters/learning/filters.py
pwnage101: It's true that `{Login,Registration}FormTPAOverridesRequested` operate similarly, but they are installed at the platform in different locations and have different pipeline step implementations. It's basically a coincidence they have the same signature.  I _could_ split out this function into a superclass, but I'd argue that is semantically poor because two separate call sites would map to a single run_filter() implementation which locks both code paths to the same inputs/outputs which just happen to coincidentally be the same today.

pwnage101:
@felipemontoya or @feanil could use one of your eyes on this big one.

The full ADR may be the best way to catch up, and there's a diagram in there too: https://github.com/openedx/edx-enterprise/blob/abd13d1be582e5c51f57373608d67032c96d4056/docs/decisions/0016-logistration-filters.md

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
         """
         data = super().run_pipeline(user=user, course_mode_data=course_mode_data, price=price)
         return data["user"], data["course_mode_data"], data["price"]
+
+
+class LogistrationContextRequested(OpenEdxPublicFilter):
```
pwnage101: Note to self: this should probably have "Legacy" in the class name to emphasize that it does not apply to the AuthN MFE logistration page.

felipemontoya: One thing that I always thought filters could help us solve is the difference between legacy and new mechanisms. Not only scope to MFE or auth, but in general.

Would it be possible to reuse this filter in the same place where the logistration MFE context is being calculated?

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+    Filter Type:
+        org.openedx.learning.auth.post_login.redirect_url.requested.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: openedx/core/djangoapps/user_authn/views/login.py
+        - Function or Method: login_user
+    """
+
+    filter_type = "org.openedx.learning.auth.post_login.redirect_url.requested.v1"
```
felipemontoya: This will be the only filter type that has `auth` in it. 

To keep it more in line with the existing filter I would call it something like this:
"org.openedx.learning.student.login.alt_redirect_url.requested.v1"

And updating the class to match `LoginAltRedirectURLRequested`. I'm calling it Alt for Alternative since any step can do it, it might not be necessarily be at the end or post-login.

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+            next_url (str): the next URL parameter from the login request.
+
+        Returns:
+            tuple[str, User, str]: the (possibly modified) redirect URL, the user, and the next URL.
+        """
+        data = super().run_pipeline(redirect_url=redirect_url, user=user, next_url=next_url)
+        return data["redirect_url"], data["user"], data["next_url"]
+
+
+class FormDescriptionProtocol(Protocol):
```
felipemontoya: This is solid work, but does not belong to the filters.py file.

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+        field_type: str = ...,
+        label: str = ...,
+        instructions: str = ...,
+        restrictions: dict = ...,
+    ) -> None:
+        """Override the given properties of the named form field."""
+        ...  # pylint: disable=unnecessary-ellipsis
+
+
+class ProviderConfigProtocol(Protocol):
```
felipemontoya: also does not belong in this file. We could move them both into a protocols.py right next to filters.py.

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+    Filter Type:
+        org.openedx.learning.logistration.response.rendered.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: openedx/core/djangoapps/user_authn/views/login_form.py
+        - Function or Method: login_and_registration_form
+    """
+
+    filter_type = "org.openedx.learning.logistration.response.rendered.v1"
```
felipemontoya: To keep things consistent with other filters and because the action there is not rendered yet, we have used completed in the past.

how about:
"org.openedx.learning.logistration.render.completed.v1" with  "LogistrationRenderCompleted"

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,250 @@ def run_filter(
...
+        """Return the registration form field values prefilled by the provider."""
+        ...  # pylint: disable=unnecessary-ellipsis
+
+
+class LoginFormTPAOverridesRequested(OpenEdxPublicFilter):
+    """
+    Filter used to apply third-party-auth overrides to the login form description.
+
+    Purpose:
+        This filter is only triggered when third-party auth is enabled and an auth pipeline
```
felipemontoya: Is it necessary that we only trigger this for third party auth and never for non third party auth? I think we could make the case here and already make this override capability functional for the general case. Enterprise/TPA don't lose anything, but any other extension cases gain a lot.

I researched a bit with claude to see how much of a detour that would be.

Result:
Each form description has exactly one construction point:

- get_login_session_form(request) — login_form.py:77
- RegistrationFormFactory.get_registration_form(request) — registration_form.py:567

and each is called from exactly two places: the public API endpoint (login.py:756, register.py:557) and the logistration page (login_form.py:302-303). One line in each function covers 100% of form builds, including every case the TPA-scoped version covers, for the same one-line modification count.

We know the need is real: PROFILE_EXTENSION_FORM (registration_form.py:337, with REGISTRATION_EXTENSION_FORM deprecated at :342) exists precisely so operators can extend the registration form. It's settings-based, requires shipping a Django Form class, and has no login-side equivalent. A general filter is the natural modern successor to that mechanism. A TPA-scoped misses the opportunity, which means a second, overlapping filter at the same code point later.

openedx_filters/learning/filters.py
felipemontoya: Researched a bit and it looks possible:

MFEContextView.get (user_authn/api/views.py:50) builds a context dict via get_mfe_context(request, redirect_to, third_party_auth_hint) at :73 — structurally the same shape as login_and_registration_form's context. So it's one filter with two trigger points, and documenting-filters-classes.rst explicitly supports listing multiple triggers in the docstring. That also answers the author's own note-to-self at filters.py:1941 ("should probably have Legacy in the class name") — if the filter covers both, it shouldn't.

felipemontoya (changes requested):
I think this work is good and overall very needed in the login/registration extensibility story. Thanks for putting this together @pwnage101. Now that I understand the context, new review should be quicker.

openedx_filters/learning/filters.py
pwnage101: Interesting, sounds promising.

pwnage101: After looking into this myself, I disagree about the context shape being structurally compatible between the two hook points.  The keys inside the context dict are nested differently, and furthermore, the enterprise pipeline logic for each is also different.  This is the only feedback of yours that I discarded.

pwnage101:
@felipemontoya Awesome feedback!  I just updated this PR to incorporate all of your feedback, with the exception of one:

I opted not to create a consolidated ContextGenerated filter.  The reality of the two "context" dictionaries is that the shape differs between the legacy template context and the AuthN MFE context, so their processing cannot be safely collapsed under one filter.  With extra work on the MFE, we could bring the two context dicts into alignment, but that is out of scope.

But, that did make me realize that I was completely missing an `AuthnMFEContextGenerated` filter, so I added that.

Additionally I created a new architecture subdomain called "Authentication", and added a new ADR (0008) to justify it.

Here's a quick overview of the 6 new authentication filters, plus the 2 pre-existing ones which still live under the "Learning" subdomain: 

| Class | Type |
|---|---|
| `LogistrationViewContextGenerated` | `org.openedx.authentication.logistration_view.context.generated.v1` |
| `LogistrationViewRenderCompleted` | `org.openedx.authentication.logistration_view.render.completed.v1` |
| `AuthnMFEContextGenerated` | `org.openedx.authentication.mfe.context.generated.v1` |
| `LoginFormGenerated` | `org.openedx.authentication.login.form.generated.v1` |
| `RegistrationFormGenerated` | `org.openedx.authentication.registration.form.generated.v1` |
| `LoginAltRedirectURLRequested` | `org.openedx.authentication.login.alt_redirect_url.requested.v1` |
| `StudentRegistrationRequested` | `org.openedx.learning.student.registration.requested.v1` |
| `StudentLoginRequested` | `org.openedx.learning.student.login.requested.v1` |

For the platform reference, you'll need to refer to https://github.com/edx/edx-platform/pull/407

openedx_filters/learning/filters.py
pwnage101: adopted this proposal.

docs/conf.py
```diff
@@ -37,13 +37,19 @@
     'sphinx_copybutton',
     'sphinx.ext.graphviz',
     'sphinxcontrib.mermaid',
+    'myst_parser',
```
pwnage101, line 40: markdown support for docs is needed because on Github only markdown supports mermaid rendering.

openedx_filters/authentication/protocols.py
```diff
@@ -0,0 +1,47 @@
...
+
+class FormDescriptionProtocol(Protocol):
+    """
+    Structural interface of the FormDescription object passed to the form-override filters.
+
+    Only the minimal surface consumed by pipeline steps is declared here. Pipeline steps
+    should not rely on anything beyond this protocol.
+    """
+
+    def override_field_properties(
```
felipemontoya: could one use this to add a new field?

My understanding is that it would not work because it only works on existing fields, but adding the capacity to add a filter to the protocol would be a killer feature for anyone trying to have their own logistration logic.

docs/decisions/0008-authentication-subdomain.md
```diff
@@ -0,0 +1,100 @@
...
+being introduced: hooks around the login/registration page context and render
+lifecycle, the login and registration form descriptions, the authentication MFE
+(`frontend-app-authn`) context, and the post-login redirect. These needed a
+home, and `learning` is a poor fit: authentication gates access for *every* user
+of the platform — learners and content authors alike — so it is a distinct
+bounded context rather than a learning activity.
+
+## Decision
+
+We will introduce a new "Authentication" architecture subdomain, implemented in
```
felipemontoya, line 26: @mariajgrimaldi this makes sense to me, but I'll like to know your thoughs. Should this mean that we also add a similar subdomain in events?

openedx_filters/authentication/protocols.py
pwnage101: Indeed that sounds like it could be really useful, but in reality just adding `add_field()` support to the Protocol would not be sufficient because it would not automatically update the POST endpoint to ingest the new field or automatically update any django models to persist the new field.  Furthermore, it offers no controls over field ordering.

We already have a feature to customize the registration forms via `PROFILE_EXTENSION_FORM` (or the deprecated `REGISTRATION_EXTENSION_FORM`) which already supports all of the above:

1. adding new fields,
2. automatically ingesting them from POST,
3. automatically persisting them in the DB,
4. providing a mechanism for field ordering.

The main downside is that these are static settings loaded once at startup so they cannot be dynamic or sensitive to TPA state or other per-user conditions, hence the need for this openedx-filter.  (Also, there's no way to add custom login fields, only registration)

The good news is that we can have the best of both worlds by leaning on `PROFILE_EXTENSION_FORM` to *add* fields, and these `*FormGenerated` filters to *dynamically hide* fields.  I'd advocate we keep the Protocol limited to override-only, and document why it is limited and point readers to `PROFILE_EXTENSION_FORM` for adding fields.

pwnage101: For now, I updated the docstrings to explain these details.

ssurendrannair (approved):
Approved — Looks good to me. 
One suggestion for follow-up: consider adding negative-path tests for malformed or unexpected filter outputs. Since these filters are extension points in the authentication flow, validating failure scenarios would strengthen the contract and help catch regressions as new implementations are added.

felipemontoya (approved):
Thanks for all the work @pwnage101 writing and refactoring after review. I think this PR has reached a very good form now.

setup.py
```diff
@@ -145,6 +145,7 @@ def is_requirement(line):
     packages=[
         "openedx_filters",
     ],
+    package_data={"openedx_filters": ["py.typed"]},
```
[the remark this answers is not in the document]
pwnage101: this seems wrong, the built package is fine, I didn't touch this list.

PR #344, opened 2026-03-27
feat: add contextmanager-based management command observability filter

ktyagiapphelix2u:
## Summary

This PR introduces the ManagementCommandContextmanagerRequested filter under the openedx_filters.management package.

The filter is invoked before a Django management command is executed, allowing plugins to inspect or modify the context manager, command name, and service variant through the standard Open edX filter pipeline. It follows the conventional filter contract where the inputs map directly to the returned tuple.

This PR also includes unit tests verifying the filter returns the expected values when no pipeline steps modify the inputs.

## Changes
Added the openedx_filters.management package.
Added the ManagementCommandContextmanagerRequested public filter.
Added unit tests for the filter's default behavior.

### DIFF
--- a/openedx_filters/management/filters.py
+++ b/openedx_filters/management/filters.py
@@ -1,0 +1,53 @@
+"""
+Package where filters related to management command execution are implemented.
+"""
+
+from collections.abc import Callable
+from typing import Any
+
+from openedx_filters.tooling import OpenEdxPublicFilter
+
+
+class ManagementCommandExecutionRequested(OpenEdxPublicFilter):
+    """
+    Filter used to modify the execution of Django management commands.
+
+    Purpose:
+        This filter is triggered in ``manage.py`` before a management command is
+        executed, allowing pipeline steps to wrap or replace the command runner.
+
+    Filter Type:
+        org.openedx.platform.management.command.execute.requested.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: manage.py
+        - Function or Method: __main__
+    """
+
+    filter_type = "org.openedx.platform.management.command.execute.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        command_name: str,
+        service_variant: str,
+        command_runner: Callable[..., Any],
+    ) -> dict[str, Any]:
+        """
+        Process management command execution arguments through the pipeline.
+
+        Arguments:
+            command_name (str): name of the management command being executed.
+            service_variant (str): service variant, such as lms or cms.
+            command_runner (Callable): callable that executes the command.
+
+        Returns:
+            dict[str, Any]: accumulated pipeline output, including possibly
+                modified command metadata and command runner.
+        """
+        return super().run_pipeline(
+            command_name=command_name,
+            service_variant=service_variant,
+            command_runner=command_runner,
+        )
--- a/openedx_filters/management/filters.py
+++ b/openedx_filters/management/filters.py
@@ -19,7 +19,7 @@ class ManagementCommandContextmanagerRequested(OpenEdxPublicFilter):
         org.openedx.platform.management.command.contextmanager.requested.v1
 
     Trigger:
-        - Repository: edx/edx-platform
+        - Repository: openedx/openedx-platform
         - Path: manage.py
         - Function or Method: __main__
     """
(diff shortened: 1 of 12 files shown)


openedx_filters/management/filters.py
```diff
@@ -0,0 +1,53 @@
...
+
+    filter_type = "org.openedx.platform.management.command.execute.requested.v1"
+
+    @classmethod
+    def run_filter(
+        cls,
+        command_name: str,
+        service_variant: str,
+        command_runner: Callable[..., Any],
+    ) -> dict[str, Any]:
```
pwnage101: This signature is non-conventional.  The input args (excluding cls) should map to an output tuple with identical parts:

```suggestion
    @classmethod
    def run_filter(
        cls,
        command_name: str,
        service_variant: str,
        command_runner: Callable[..., Any],
    ) -> tuple[str, str, Callable[..., Any]]:
```

But also, see my comment on your platform PR:  https://github.com/edx/edx-platform/pull/200#discussion_r3659753568

The revised signature would look like this instead:

```suggestion
    @classmethod
    def run_filter(
        cls,
        command_contextmanager: AbstractContextManager[None]
        command_name: str,
        service_variant: str,
    ) -> tuple[AbstractContextManager[None], str, str]:
```

pwnage101 (changes requested):

openedx_filters/management/filters.py
ktyagiapphelix2u: I updated this filter to the context-manager pattern so inputs and outputs map positionally as a tuple. The signature is now command_contextmanager, command_name, service_variant -> (command_contextmanager, command_name, service_variant), and the test was updated to assert tuple unpacking accordingly. This also aligns with the platform-side change where manage.py remains the only execution callsite and the filter only wraps execution via a context manager.

openedx_filters/management/filters.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Purpose:
+        This filter is triggered in ``manage.py`` before a management command is
+        executed, allowing pipeline steps to provide a context manager wrapper.
+
+    Filter Type:
+        org.openedx.platform.management.command.contextmanager.requested.v1
+
+    Trigger:
+        - Repository: edx/edx-platform
```
feanil:
```suggestion
        - Repository: openedx/openedx-platform
```

feanil (commented):
The description currently describes what this is but it's unclear why we would want this filter? What's the reason to inject data into a management command at runtime instead of just adding more parameters to the management command to allow for variation?  More context on the why of this filter would be useful.

Also, it would be nice to have some docs even if it's in the PR description of what it would look like to install this filter on a management command.

openedx_filters/management/tests/test_filters.py
```diff
@@ -0,0 +1,39 @@
...
+            filtered_service_variant,
+        ) = ManagementCommandContextmanagerRequested.run_filter(
+            command_contextmanager=command_contextmanager,
+            command_name=command_name,
+            service_variant=service_variant,
+        )
+
+        self.assertEqual(command_contextmanager, filtered_command_contextmanager)
+        self.assertEqual(command_name, filtered_command_name)
+        self.assertEqual(service_variant, filtered_service_variant)
```
pwnage101: Use modern pytest-style assertions:
```suggestion
        assert command_contextmanager == filtered_command_contextmanager
        assert command_name == filtered_command_name
        assert service_variant == filtered_service_variant
```

pwnage101 (commented):
Just a few more things:
- Address Feanil's comment by providing a more detailed docstring about the expected use case.
- One minor test issue (see below).
- Bump the version.
- squash all commits into 1.

openedx_filters/management/filters.py
pwnage101: +1

ktyagiapphelix2u: Updated

openedx_filters/management/tests/test_filters.py
ktyagiapphelix2u: Updated

ktyagiapphelix2u:
> The description currently describes what this is but it's unclear why we would want this filter? What's the reason to inject data into a management command at runtime instead of just adding more parameters to the management command to allow for variation? More context on the why of this filter would be useful.
> 
> Also, it would be nice to have some docs even if it's in the PR description of what it would look like to install this filter on a management command.

This filter isn't used to send extra options to management commands. Instead, it runs every management command in the same shared setup.

Management commands don't go through Django's normal web request process, so the usual middleware doesn't run. This filter gives you one place to add things like logging, performance monitoring, error tracking, or Datadog tracing for all management commands, without having to change each command separately.

PR #349, opened 2026-04-08
build: Remove outdated sphinx-contentui package

sarina:
See https://github.com/openedx/docs.openedx.org/pull/1417

This is an outdated package that was included in the docs cookie cutter. It can cause build errors and doesn't seem to be used.


mariajgrimaldi (approved):

mariajgrimaldi:
Thanks a lot!

PR #351, opened 2026-04-17
fix: expand Python version compatibility

kiram15:
## Description

This PR adds the flexibility to support Python 3.11 as well as 3.12 to unblock work [to transition all enterprise code into well-defined plugin hooks](https://discuss.openedx.org/t/rfc-pluginifying-edx-enterprise-looking-for-community-interest-collaborators/18316). 

## Motivation:
Right now, the repo works well with both python 3.11 and 3.12 ([link](https://github.com/pwnage101/openedx-filters/blob/ccf780ff2fa9501858b0320eda9b2cbbc538ed6d/.github/workflows/ci.yml#L18)), but the constraint in the setup file deems that only versions past 3.12 are allowed.

This blocks all of the work that the Titans team (enterprise team at 2u)  is doing to deploy new plugins, as they have not yet migrated to 3.12 in edx/edx-platform. This in turn is slowing down the work of removing the hard dependency of the platform code in the edx-enterprise repos.

There is no specific functionality that is being used from python 3.12, so this would unblock that initiative.

## Supporting information

To speed this up, some maintainers were pinged via slack, but they are also asked as reviewers here.

The proposal here is to allow installers to use in 3.11 at their own risk. CI testing would remain using 3.12 only.

## Testing instructions

It is possible to run the test suite using 3.11.

``` bash
  python3.11 -m venv .venv311
  source .venv311/bin/activate                                                                                                                                                             
  pip install -e ".[dev]"   # or pip install -e . && pip install -r requirements/test.txt
  python -m pytest                                                                                                                                                                         
  Expected: all tests pass without syntax errors or import failures.
```

## Deadline

Soon if possible to speed up the work of removing enterprise.

## Checklists

Check off if complete *or* not applicable:

**Merge Checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Changelog entry added using scriv with short description of the change

**Post Merge:**
- [ ] Trigger the release workflow to create a new GitHub release.
- [ ] Check new version is pushed to PyPI after tag-triggered build is finished.

### DIFF
--- a/openedx_filters/__init__.py
+++ b/openedx_filters/__init__.py
@@ -1,6 +1,17 @@
 """
 Filters of the Open edX platform.
 """
+import sys
+import warnings
+
 from openedx_filters.filters import *
 
-__version__ = "3.0.0"
+__version__ = "3.3.0"
+
+if sys.version_info < (3, 12):
+    warnings.warn(
+        "Python 3.11 support is deprecated and will be removed in a future release. "
+        "Please upgrade to Python 3.12 or later.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,14 @@ jobs:
         os: [ubuntu-latest]
         python-version: ['3.12']
         toxenv: ["django42", "django52", "quality", "docs"]
+        # [TEMP] include 3.11 tests until most operators developing filters have migrated to Verawood/Python 3.12.
+        include:
+          - os: ubuntu-latest
+            python-version: '3.11'
+            toxenv: "py311-django42"
+          - os: ubuntu-latest
+            python-version: '3.11'
+            toxenv: "py311-django52"
 
     steps:
     - uses: actions/checkout@v6
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,8 +32,6 @@ Change Log
 Unreleased
 ----------
 
-[2.1.0] - 2025-04-23
-
 Added
 ~~~~~
 
@@ -44,6 +42,13 @@ See the fragment files in the `changelog.d directory`_.
 .. _changelog.d directory: https://github.com/openedx/openedx-filters/tree/master/changelog.d
 .. scriv-insert-here
 
+[3.3.0] - 2025-04-17
+--------------------
+
+Changed
+~~~~~~~
+
+* Expand Python version compatibility
 
 [2.0.1] - 2025-02-18
 --------------------
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,7 @@ setup(
     ],
     include_package_data=True,
     install_requires=load_requirements("requirements/base.in"),
-    python_requires=">=3.12",
+    python_requires=">=3.11",
     license="AGPL 3.0",
     zip_safe=False,
     keywords="Python openedx",


feanil (commented):
This seems fine to me to help with the existing transition but if we're saying we support both 3.11 and 3.12 we should test with both.  However, there is some complexity around the requirements workflows, because the library pins requirements for testing but not for the published version, the verified working versions of the requirements actually lives in the openedx/openedx-platform repo.  If you want to pursue this, you should add a workflow to build requirements for both 3.11 and 3.12 and then test with the relevant files for the relevant test jobs.  This won't be perfect but it would catch issues before you try to deploy.

Alternatively, this is risk that the 2U org can take on and just be okay knowing that the latest version of the library might have dependency mismatches for 3rd party libraries.

pwnage101:
Notes from our standup discussion:
* We originally wanted to just accept the risk because openedx-filters is a relatively simple repository with minimal surface area for exposure to 3.12 breaking changes.
* That said, we're open to adding a 3.11 unit test step but concerned about the amount of work vs. available capacity. Maybe it's as easy as adding a test runner that runs make upgrade under 3.11 and runs tests, but we just don't know for sure if there will be any snags, and if it will be sensitive to unexpected changes to common_constraints.txt.

pwnage101:
@kiram15 @feanil this patch works for me. Test using this command: `tox -e py311-django52` and watch it generate new requirements files, install them, and tests pass.

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index ee33958..2497978 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,14 @@ jobs:
         os: [ubuntu-latest]
         python-version: ['3.12']
         toxenv: ["django42", "django52", "quality", "docs"]
+        # [TEMP] include 3.11 tests until most operators developing filters have migrated to Verawood/Python 3.12.
+        include:
+          - os: ubuntu-latest
+            python-version: '3.11'
+            toxenv: "py311-django42"
+          - os: ubuntu-latest
+            python-version: '3.11'
+            toxenv: "py311-django52"
 
     steps:
     - uses: actions/checkout@v6
diff --git a/tox.ini b/tox.ini
index b368e26..3e78f0e 100644
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py312-django{42,52}, quality, docs
+envlist = py{311,312}-django{42,52}, quality, docs
 
 [doc8]
 ; D001 = Line too long
@@ -16,10 +16,15 @@ addopts = --cov openedx_filters --cov-report term-missing --cov-report xml
 norecursedirs = .* docs requirements site-packages
 
 [testenv]
+allowlist_externals =
+    make
 deps =
     django42: Django>=4.2,<4.3
     django52: Django>=5.2,<5.3
-    -r{toxinidir}/requirements/test.txt
+    !py311: -r{toxinidir}/requirements/test.txt
+commands_pre =
+    py311: make upgrade
+    py311: pip install -r{toxinidir}/requirements/test.txt
 commands =
     pytest {posargs}
```

openedx_filters/__init__.py
```diff
@@ -1,6 +1,17 @@
...
-__version__ = "3.2.0"
+__version__ = "3.3.0"
+
+if sys.version_info < (3, 12):
+    warnings.warn(
+        "Python 3.11 support is deprecated and will be removed in a future release. "
+        "Please upgrade to Python 3.12 or later.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
deborahgu: smart to add this.

deborahgu (approved):
lgtm, assuming everything still works.

feanil:
@pwnage101 that approach seems fine to me as a temporary solution to this issue.

felipemontoya (approved):
I think the approach is good and having the deprecation notice is a good move.

Before we merge, we would need to solve the issues with the codecov. Probably by ignoring the py311 run.

I leave my approve to make the process expedite.

PR #363, opened 2026-05-28
feat: adding CourseEnrollmentViewStarted to execute in enrollment view

kiram15:
## Description

This adds a filter `CourseEnrollmentViewStarted` that will be emitted by the enterprise enrollment view before enrollment is created. We are not reusing the existing `CourseEnrollmentStarted` filter because this is triggered at the model level, and we want to scoped this to the view-layer enrollment path, not to every programmatic enrollment. 

This is part of an ongoing initiative to leverage openedx-filters to allow enterprise to customize logic and remove it from edx-platform code. 

## Supporting information

Jira ticket - [ENT-11570](https://2u-internal.atlassian.net/browse/ENT-11570)

## Testing instructions

See testing instructions in [edx-enterprise PR](https://github.com/openedx/edx-enterprise/pull/2553) 

## Deadline

"None" if there's no rush, or provide a specific date or event (and reason) if there is one.

## Other information

Include anything else that will help reviewers and consumers understand the change.

- Any other PRs or issues that should be linked here? Any related PRs?
- Any special concerns or limitations? For example: deprecations, security, or anything you think should be noted.

## Checklists

Check off if complete *or* not applicable:

**Merge Checklist:**
- [ ] All reviewers approved
- [ ] Reviewer tested the code following the testing instructions
- [ ] CI build is green
- [ ] Changelog entry added using scriv with short description of the change
- [ ] Code dependencies reviewed
- [ ] Fixup commits are squashed away
- [ ] Unit tests added/updated
- [ ] Noted any: Concerns, dependencies, migration issues, deadlines, tickets

**Post Merge:**
- [ ] Trigger the release workflow to create a new GitHub release.
- [ ] Check new version is pushed to PyPI after tag-triggered build is finished.
- [ ] Upgrade the package in the Open edX platform requirements (if applicable)

### DIFF
--- a/openedx_filters/__init__.py
+++ b/openedx_filters/__init__.py
@@ -6,7 +6,7 @@ import warnings
 
 from openedx_filters.filters import *
 
-__version__ = "3.5.0"
+__version__ = "3.6.0"
 
 if sys.version_info < (3, 12):  # pragma: no cover
     warnings.warn(
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -341,6 +341,62 @@ class CourseUnenrollmentStarted(OpenEdxPublicFilter):
         return data.get("enrollment")
 
 
+class CourseEnrollmentViewStarted(OpenEdxPublicFilter):
+    """
+    Filter used to perform pre-enrollment processing during the enrollment REST API view.
+
+    Purpose:
+        This filter is triggered when a user initiates enrollment via the enrollment REST API view,
+        just before the enrollment is created, allowing pipeline steps to perform pre-enrollment
+        processing scoped to the view layer.
+
+    Filter Type:
+        org.openedx.learning.course.enrollment.view.started.v1
+
+    Trigger:
+        - Repository: openedx/openedx-platform
+        - Path: openedx/core/djangoapps/enrollments/views.py
+        - Function or Method: EnrollmentListView.post
+    """
+
+    filter_type = "org.openedx.learning.course.enrollment.view.started.v1"
+
+    class PreventEnrollment(OpenEdxFilterException):
+        """
+        Raise to prevent the enrollment process from continuing.
+        """
+
+    @classmethod
+    def run_filter(
+        cls, user: Any, course_key: CourseKey, requester_is_backend_service: bool
+    ) -> tuple[Any, CourseKey, bool]:
+        """
+        Process the user, course_key, and requester_is_backend_service using the configured
+        pipeline steps to preempt the enrollment process.
+
+        Arguments:
+            user (User): Django User enrolling in the course.
+            course_key (CourseKey): course key associated with the enrollment.
+            requester_is_backend_service (bool): if request was made by a server with an API key.
+
+        Returns:
+            tuple[Any, CourseKey, bool]:
+                - User: Django User object.
+                - CourseKey: course key associated with the enrollment.
+                - bool: if request was made by a server with an API key.
+        """
+        data = super().run_pipeline(
+            user=user,
+            course_key=course_key,
+            requester_is_backend_service=requester_is_backend_service
+        )
+        return (
+            data["user"],
+            data["course_key"],
+            data["requester_is_backend_service"],
+        )
+
+
 class CertificateCreationRequested(OpenEdxPublicFilter):
     """
     Filter used to modify the certificate creation process for a given user in a course.
--- a/openedx_filters/learning/tests/test_filters.py
+++ b/openedx_filters/learning/tests/test_filters.py
@@ -21,6 +21,7 @@ from openedx_filters.learning.filters import (
     CourseEnrollmentAPIRenderStarted,
     CourseEnrollmentQuerysetRequested,
     CourseEnrollmentStarted,
+    CourseEnrollmentViewStarted,
     CourseHomeUrlCreationStarted,
     CourseRunAPIRenderStarted,
     CourseStartDateValidationFailed,
@@ -77,10 +78,7 @@ class TestCertificateFilters(TestCase):
             generation_mode,
         )
 
-        self.assertTupleEqual(
-            (user, course_key, mode, status, grade, generation_mode,),
-            result,
-        )
+        assert result == (user, course_key, mode, status, grade, generation_mode,)
 
     def test_certificate_render_started(self):
         """
@@ -97,7 +95,7 @@ class TestCertificateFilters(TestCase):
 
         result = CertificateRenderStarted.run_filter(context, template_name)
 
-        self.assertTupleEqual((context, template_name,), result)
+        assert result == (context, template_name,)
 
     @data(
         (CertificateRenderStarted.RedirectToPage, {"redirect_to": "custom-certificate.pdf"}),
@@ -115,7 +113,7 @@ class TestCertificateFilters(TestCase):
         """
         exception = CertificateException(message="You can't generate certificate", **attributes)
 
-        self.assertLessEqual(attributes.items(), exception.__dict__.items())
+        assert attributes.items() <= exception.__dict__.items()
 
 
 @ddt
@@ -144,7 +142,7 @@ class TestAuthFilters(TestCase):
 
         form_data = StudentRegistrationRequested.run_filter(expected_form_data)
 
-        self.assertEqual(expected_form_data, form_data)
+        assert expected_form_data == form_data
 
     @patch(
         "openedx_filters.tooling.OpenEdxPublicFilter.run_pipeline",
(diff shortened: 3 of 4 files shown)


openedx_filters/learning/filters.py
```diff
@@ -340,6 +340,66 @@ def run_filter(cls, enrollment: Any) -> Any:
...
+    Purpose:
+        This filter is triggered when a user initiates enrollment via the enrollment REST API view,
+        just before the enrollment is created, allowing pipeline steps to perform pre-enrollment
+        processing scoped to the view layer.
+
+    Filter Type:
+        org.openedx.learning.course.enrollment.view.started.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
```
```suggestion
        - Repository: openedx/openedx-platform
```

openedx_filters/learning/filters.py
```diff
@@ -340,6 +340,66 @@ def run_filter(cls, enrollment: Any) -> Any:
...
+    filter_type = "org.openedx.learning.course.enrollment.view.started.v1"
+
+    class PreventEnrollment(OpenEdxFilterException):
+        """
+        Raise to prevent the enrollment process from continuing.
+        """
+
+    @classmethod
+    def run_filter(
+        cls, user: Any, course_key: CourseKey, linked_enterprise: (str | None), has_api_key_permissions: bool
```
pwnage101: ❌ "enterprise" appears in the filter signature, which is deal-breaking because we're trying to remove all enterprise references from the platform. ONLY the enterprise pipeline step logic should handle fetching the linked enterprise from the request data (using `crum`).

pwnage101 (changes requested):

openedx_filters/learning/tests/test_filters.py
```diff
@@ -312,6 +330,24 @@ def test_halt_queryset_request(self, request_exception, attributes):
 
         self.assertLessEqual(attributes.items(), exception.__dict__.items())
 
+    @data(
+        (
+            CourseEnrollmentViewStarted.PreventEnrollment,
+            {"message": "Enterprise enrollment processing failed"},
+        ),
+    )
```
pwnage101: There's no reason for this to be ddt driven if there's only one case.  Should be a simple fix (hard code the values in the test).

openedx_filters/learning/tests/test_filters.py
```diff
@@ -312,6 +330,24 @@ def test_halt_queryset_request(self, request_exception, attributes):
...
+        """
+        Test CourseEnrollmentViewStarted.PreventEnrollment exception attributes.
+
+        Expected behavior:
+            - The exception must carry the message attribute specified.
+        """
+        exception = enrollment_exception(**attributes)
+
+        self.assertLessEqual(attributes.items(), exception.__dict__.items())
+
```
pwnage101: Two things here:

1. Nit: Please use `assert x <= y` instead of `self.assertLessEqual(x, y)`

Background: We've been trying to migrate away from `self.assert...()` helper functions since about 2018, and we've struggled to finish the migration ever since.  Now, openedx-platform and axim are more aggressively linting to block new `self.assert...()` calls from being added, but that linting only runs in platform CI currently.

2. Nit: This test is borderline useless.  I think we really care that a constructed exception has expected attributes available.  Something like:
```suggestion
        test_message = "Enterprise enrollment processing failed"
        exception = CourseEnrollmentViewStarted.PreventEnrollment(message=test_message)
        assert exception.message == test_message
```

openedx_filters/learning/filters.py
```diff
@@ -340,6 +340,62 @@ def run_filter(cls, enrollment: Any) -> Any:
...
+        data = super().run_pipeline(
+            user=user,
+            course_key=course_key,
+            has_api_key_permissions=has_api_key_permissions
+        )
+        return (
+            data.get("user"),
+            data.get("course_key"),
+            data.get("has_api_key_permissions"),
+        )
```
pwnage101, line 397: ❌ This breaks type checking.  Proof:

```
$ mypy openedx_filters/learning/filters.py
openedx_filters/learning/filters.py:392: error: Incompatible return value type (got "tuple[Any | None, Any | None, Any | None]", expected "tuple[Any, CourseKey, bool]")  [return-value]
Found 1 error in 1 file (checked 1 source file)
```

the function signature has a return value `-> tuple[Any, CourseKey, bool]` but the actual return value is `-> tuple[Any, CourseKey | None, bool | None]` due to using `data.get()` instead of `data[]`.

Fix:
```suggestion
        return (
            data["user"],
            data["course_key"],
            data["has_api_key_permissions"],
        )
```

pwnage101 (changes requested):
Few more changes needed.

pwnage101 (approved):
LGTM. @felipemontoya could you take a look at this one too?

openedx_filters/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 from openedx_filters.filters import *
 
-__version__ = "3.4.1"
+__version__ = "3.5.0"
```
felipemontoya: we are already at 3.5.0, so this would have to be updated to 3.6.0

openedx_filters/learning/filters.py
```diff
@@ -340,6 +340,62 @@ def run_filter(cls, enrollment: Any) -> Any:
         return data.get("enrollment")
 
 
+class CourseEnrollmentViewStarted(OpenEdxPublicFilter):
```
felipemontoya, line 344: The filter as it is, looks well crafted and correct.

Now, the purpose of the filter:

```
    Purpose:
        This filter is triggered when a user initiates enrollment via the enrollment REST API view,
        just before the enrollment is created, allowing pipeline steps to perform pre-enrollment
        processing scoped to the view layer.
```

Is very similar to the already existing `org.openedx.learning.course.enrollment.started.v1`. One operates in the view (the new one) and the other operates in the enrollment API.

As a maintainer I have to check, have you considered using the existing filter? I know you are going for a 1:1 replacement of enterprise, but adding a new filter is a promise we are making long term for every developer that jumps in the wagon.

felipemontoya (commented):
The code looks good. There is a question about the possibility of using the existing filter that operates right before an enrollment.

openedx_filters/learning/filters.py
kiram15: I was going to use the existing CourseEnrollmentStarted hook, but that is tied to the model-layer enrollment flow, not the legacy view-layer enterprise flow. Attaching enterprise side-effects (posting to the enterprise enrollment API, recording data sharing consent) to CourseEnrollmentStarted would cause those calls to fire on every programmatic enrollment that happens (API calls, management commands, bulk enrollment scripts, test fixtures, signals, etc) to involve an enterprise-linked user, including internal tooling and migrations that have no awareness of or need for enterprise processing.

openedx_filters/learning/tests/test_filters.py
```diff
@@ -268,6 +270,22 @@ def test_course_unenrollment_started(self):
...
+            - The filter must have the signature specified.
+            - The filter should return user, course_key, and has_api_key_permissions, in that order.
+        """
+        user = Mock()
+        course_key = Mock()
+        has_api_key_permissions = True
+
+        result = CourseEnrollmentViewStarted.run_filter(user, course_key, has_api_key_permissions)
+
+        self.assertTupleEqual((user, course_key, has_api_key_permissions), result)
```
pwnage101: use pytest-style `assert` statements instead of unittest-style `self.assert*` methods.

openedx_filters/learning/filters.py
pwnage101: Yes, this took a lot of deliberation, but we ultimately landed on creating a new filter to cover this distinct case.  The enrollment view is starting, but NOT the lower-level enrollment.

openedx_filters/learning/filters.py
```diff
@@ -341,6 +341,62 @@ def run_filter(cls, enrollment: Any) -> Any:
...
+        Returns:
+            tuple[Any, CourseKey, bool]:
+                - User: Django User object.
+                - CourseKey: course key associated with the enrollment.
+                - bool: if request was made by a server with an API key.
+        """
+        data = super().run_pipeline(
+            user=user,
+            course_key=course_key,
+            requester_is_backend_service=requester_is_backend_service
```
pwnage101, line 391: Not sure why CI didn't catch this, but there's a missing trailing comma

openedx_filters/learning/filters.py
felipemontoya: Got it. I understant the use case and I think it is reasonable. This was my main outstanding comment so I'll approve the PR.

felipemontoya (approved):
Now that the difference with the existing deeper filter that operates on the enrollment API is clear I leave my approval.

pwnage101:
Thanks Felipe.  Now I'm just trying to understand why CI checks (unit tests, quality tests) did not run.

Edit: CI checks are running again.

PR #381, opened 2026-07-27
feat: modernize to uv + pyproject.toml + semantic-release

irfanuddinahmad:
## Summary

Modernize `openedx-filters` to uv + pyproject.toml (PEP 621/735) + python-semantic-release.

Part of [openedx/public-engineering#506](https://github.com/openedx/public-engineering/issues/506).

- Move `openedx_filters/` to `src/openedx_filters/` (org decision on #506: src/ layout is in scope this cycle, precedent in `xblocks-extra`) — fixes editable-install resolution under tools like mypy; does **not** change the installed package's import path (verified: wheel contents are unchanged, still `openedx_filters/...` at the wheel root, no `src/` leaks in)
- Replace `setup.py`/`setup.cfg` with `pyproject.toml` (PEP 621 static metadata); `.coveragerc` and `setup.cfg`'s `[isort]` section also folded into `pyproject.toml`
- Switch from pip-compile to `uv` with PEP 735 dependency groups; commit `uv.lock`
- Update `tox.ini` to use `tox-uv` with `uv-venv-lock-runner`
- Update CI to use `astral-sh/setup-uv`; SHA-pin all actions
- Add `python-semantic-release` + `release.yml`; replace the old manual `workflow_dispatch` release process (bump-my-version + scriv github-release) and the old token-based `pypi-publish.yml`
- `commitlint.yml` already existed in this repo and needed no changes

## Not included

- Lint tooling (pylint/isort/pycodestyle/mypy/ruff) left exactly as configured before — out of scope per the parent ticket, and this repo already had all of them wired up.
- CHANGELOG.rst / `changelog.d` (scriv) is left in place as the manually-curated changelog; `python-semantic-release` is configured with `changelog: "false"` so it only owns version tagging/PyPI publish, not changelog content. `changelog.d/scriv.ini`'s `version = literal: openedx_filters/__init__.py: __version__` setting was removed since that `__version__` string no longer exists (see Versioning below) — running `make changelog` (`scriv collect`) will now just omit a version number from the entry heading, which a maintainer can pass manually if desired.
- **`.github/workflows/upgrade-python-requirements.yml` has a latent gap this PR can't fix.** It calls `openedx/.github`'s shared reusable workflow, which hardcodes `add-paths: requirements`. Now that `requirements/` is deleted, the scheduled job will silently stop producing real dependency-upgrade PRs — same gap across every repo in this migration batch; needs a fix in `openedx/.github` itself, not here.

## Versioning

Dynamic, via `setuptools-scm` from git tags (previously a hardcoded `__version__` string in `openedx_filters/__init__.py`, kept in sync by hand via `bump-my-version`). `__version__` is now read at import time via `importlib.metadata`.

Pre-flight check: the latest git tag (`v3.8.0`, via `git tag --sort=-v:refname`) matches the version currently live on PyPI (3.8.0), so the first semantic-release run has an accurate base to compute the next version from.

## Testing Notes

Verified locally:
- `uv lock` resolves cleanly (129 packages)
- `uv sync --group dev` succeeds; editable install resolves `openedx_filters.__file__` to `src/openedx_filters/__init__.py`
- `uv run mypy` succeeds against the new src/ layout (17 source files, no issues) — the specific resolution problem src/ layout is meant to fix
- Full tox matrix passes: `py311-django42`, `py311-django52`, `django42`, `django52` (110 tests each, after merging in `main`'s `SupportContactContextRequested` filter — see below), `quality` (pylint/mypy/pycodestyle/ruff/isort/build+twine check), `docs` (Sphinx build + doc8, including confirming `linkcode_resolve` now emits correct `.../blob/main/src/openedx_filters/...` URLs)
- Merged in `main` after it landed a `SupportContactContextRequested` filter (with its own tests) post-dating this branch's src-layout move; ported it into `src/`, verified its 3 tests pass and the full `quality` env stays clean
- `uv build` produces a wheel/sdist with correct metadata (version 3.8.0 from the git tag, `AGPL-3.0` SPDX license, test subpackages excluded from the wheel) and unchanged wheel contents (`openedx_filters/...` at the wheel root, byte-identical file listing to before the src/ move)

Could not be verified locally (needs CI / maintainer action):
- The actual GitHub Actions run (matrix jobs, `release.yml`'s reusable-workflow call, OIDC token exchange) — validated the YAML parses and the job/permission structure locally, but GitHub Actions-specific behavior (e.g. `secrets: inherit` through a reusable workflow call) can only be confirmed once CI runs on the PR.
- PyPI OIDC publish itself (see below).

## Code reviewer notes

- **PyPI trusted publisher (OIDC) is not yet confirmed configured for this project.** `release.yml`'s `publish_to_pypi` job assumes a PyPI trusted publisher is set up for `openedx-filters` (Settings → Publishing on the PyPI project page, pointing at `openedx/openedx-filters`, workflow `release.yml`, environment optional). This can't be configured from a PR — a maintainer with PyPI project admin access needs to confirm/add it before the first automated release can actually publish. Until then, `release`/tagging/GitHub-release-creation will still work; only the final `publish_to_pypi` step would fail.
- **`uv`/`tox-uv` version pinning:** while verifying this locally, different `uv` releases turned out to serialize auto-derived `[tool.uv].conflicts` entries (for the `django42` dependency-group conflict) in a not-always-stable order, which made `tox-uv-bare`'s internal `uv sync --locked` check fail nondeterministically depending on which `uv` version last touched the lockfile. Fixed by (a) listing all of the transitively-conflicting groups (`test`, `doc`, `quality`, `dev`) explicitly in `[tool.uv].conflicts` instead of relying on derivation, and (b) pinning the `uv` package version itself via `[tool.edx_lint].uv_constraints` to match the `version:` pin on the `astral-sh/setup-uv` step in `ci.yml`. If a future `make upgrade` bumps the locked `uv` version, bump the `setup-uv` step's `version:` input to match (see comments in `pyproject.toml`/`ci.yml`).
- The old py311-only `commands_pre` step in `tox.ini` (`make upgrade` + manual `requirements/test.txt` install) is gone; it's no longer needed since `uv.lock` drives every env's dependencies identically regardless of Python version.

---

_This PR was created with Claude Code._


openedx-webhooks:
Thanks for the pull request, @irfanuddinahmad!


This repository is currently maintained by `@openedx/committers-openedx-filters`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

.github/workflows/ci.yml
```diff
@@ -1,17 +1,18 @@
 name: Python CI
 
 on:
-  push:
```
salman2013, line 4: We do not have to remove this.

.github/workflows/ci.yml
```diff
@@ -1,17 +1,18 @@
...
+  workflow_call:
   pull_request:
     branches:
     - '**'
 
 
 jobs:
   run_tests:
-    name: tests
+    name: ${{ matrix.toxenv }}
```
salman2013: should we follow this here?
https://github.com/openedx/staff-graded-xblock/pull/394#discussion_r4014676828

pyproject.toml
```diff
@@ -1,7 +1,194 @@
...
+include_trailing_comma = true
+indent = "    "
+line_length = 120
+multi_line_output = 3
+
+[tool.semantic_release]
+build_command = "python -m pip install --upgrade build && SETUPTOOLS_SCM_PRETEND_VERSION=$NEW_VERSION python -m build"
+major_on_zero = false
+allow_zero_version = true
+tag_format = "v{version}"
```
salman2013, line 193: This is default and we do not need to mention it explicitly.
Ref: https://python-semantic-release.readthedocs.io/en/stable/configuration/configuration.html?utm_source=chatgpt.com#tag-format

salman2013 (changes requested):

.github/workflows/ci.yml
irfanuddinahmad: `release.yml` (also added in this PR) now owns the push-to-main trigger and calls this workflow via `workflow_call` (`uses: ./.github/workflows/ci.yml`), so tests still run on every push — just through that path instead of `ci.yml`'s own trigger. Re-adding `push` here directly would double-run the full matrix on every push. Matches how XBlock does it in production.

.github/workflows/ci.yml
irfanuddinahmad: Good catch, thanks for the staff-graded-xblock#394 link — confirmed the same risk here. Reverted to a static job name in aae0c52.

pyproject.toml
irfanuddinahmad: explicit is better than implicit here. Incase PSR updated the default

salman2013 (approved):

pyproject.toml
```diff
@@ -1,7 +1,194 @@
...
+    "Framework :: Django :: 5.2",
+    "Intended Audience :: Developers",
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.12",
+]
+requires-python = ">=3.11"
+dynamic = ["version"]
+dependencies = [
+    "django>=4.2,<6.0",
```
feanil: Published 3.14.0 declares a bare `Requires-Dist: django`, so capping it here means the next release blocks Django 6.0 for every consumer of openedx-filters until someone cuts a new one.  `Django<6.0` is already in `[tool.uv].constraint-dependencies`, which is where a pin that only bounds this repo's own lockfile belongs.  Drop the cap and keep `django>=4.2`.

.github/workflows/release.yml
```diff
@@ -1,63 +1,89 @@
...
-    inputs:
-      tag:
-        description: "The type of version bump"
-        required: true
-        type: choice
-        options:
-          - "major"
-          - "minor"
-          - "patch"
+  push:
```
feanil: This moves the release from `workflow_dispatch` to push, which leaves `CHANGELOG.rst:25-31` wrong on every clause.  It still tells maintainers to trigger the workflow manually from the Actions tab, and says it will bump the version and update the changelog - PSR computes the version from the commits, and `changelog: "false"` means it doesn't touch the changelog at all.  Update those instructions along with the workflow.

.github/workflows/release.yml
```diff
@@ -1,63 +1,89 @@
...
+        name: dist
+        path: dist/
+        if-no-files-found: error
+
+  publish_to_pypi:
+    runs-on: ubuntu-latest
+    needs: release
+    if: github.ref_name == 'main' && needs.release.outputs.released == 'true'
+    permissions:
+      id-token: write
```
feanil: nit: this job only downloads an artifact and publishes it, so adding `contents: read` here would make it explicitly read only.

Makefile
```diff
@@ -17,73 +17,56 @@ clean: ## remove generated byte code, coverage reports, and build artifacts
...
 	rm -fr dist/
 	rm -fr *.egg-info
 
 coverage: clean ## generate and view HTML coverage report
-	pytest --cov-report html
+	uv run pytest --cov-report html
 	$(BROWSER)htmlcov/index.html
 
 docs: ## generate Sphinx HTML documentation, including API docs
-	tox -e docs
```
feanil: `docs:` now inlines the five commands from `tox.ini`'s `[testenv:docs]`, so they become a second copy to keep in sync.  `uv run tox -e docs` keeps them in one place.  Was there a reason to inline them?

pyproject.toml
feanil: Published 3.14.0 declares a bare `Requires-Dist: django`, so capping it here means the next release blocks Django 6.0 for every consumer of openedx-filters until someone cuts a new one.  `Django<6.0` is already in `[tool.uv].constraint-dependencies`, which is where a pin that only bounds this repo's own lockfile belongs.  Drop the cap and keep `django>=4.2`.

.github/workflows/release.yml
feanil: This moves the release from `workflow_dispatch` to push, which leaves `CHANGELOG.rst:25-31` wrong on every clause.  It still tells maintainers to trigger the workflow manually from the Actions tab, and says it will bump the version and update the changelog - PSR computes the version from the commits, and `changelog: "false"` means it doesn't touch the changelog at all.  Update those instructions along with the workflow.

.github/workflows/ci.yml
```diff
@@ -28,21 +29,24 @@ jobs:
...
+      uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
       with:
+        enable-cache: true
         python-version: ${{ matrix.python-version }}
+        # Pinned to match the `uv` package version locked in uv.lock (a
+        # transitive dependency of tox-uv/tox-uv-bare). Keep this in sync
+        # whenever `make upgrade` changes that pinned version, otherwise
+        # tox-uv's internal `--locked` check can disagree with this
+        # workflow's own `uv sync`/`uv run` calls.
+        version: "0.11.32"
```
feanil: I think this is not needed.  `tox_uv` and the uv used to run CI use different python environments. I ran the CI sequence with PATH uv 0.12.18 against the locked 0.11.32 in the tox venv and `uv sync`, `uv lock --check` and `uv run tox -e quality` all passed with `uv.lock` untouched and no complaints as far as I could tell.

.github/workflows/release.yml
feanil: nit: this job only downloads an artifact and publishes it, so adding `contents: read` here would make it explicitly read only.

Makefile
feanil: `docs:` now inlines the five commands from `tox.ini`'s `[testenv:docs]`, so they become a second copy to keep in sync.  `uv run tox -e docs` keeps them in one place.  Was there a reason to inline them?

pyproject.toml
feanil: Published 3.14.0 declares a bare `Requires-Dist: django`, so capping it here means the next release blocks Django 6.0 for every consumer of openedx-filters until someone cuts a new one.  `Django<6.0` is already in `[tool.uv].constraint-dependencies`, which is where a pin that only bounds this repo's own lockfile belongs.  Drop the cap and keep `django>=4.2`.

.github/workflows/release.yml
feanil: This moves the release to `on: push` and leaves scriv without a trigger.  `Collect changelog` and `scriv github-release` are both gone, so a merge to `main` tags and publishes without anyone running `make changelog`.  Dropping `version =` from `changelog.d/scriv.ini` also means `scriv collect` now writes a bare `2026-09-24` heading where every past entry reads `[3.14.0] - 2026-09-16`.  `CHANGELOG.rst:19-31` still describes the old flow to both contributors and maintainers.

My recommendation is to drop scriv and let PSR's notes be the changelog.  `release_notes` already feeds `gh release create --notes-file`, so the releases carry generated notes today; that means deleting `changelog.d/`, `scriv` from the dev group, both Makefile targets, and the contributor instructions.  If you'd rather keep scriv, it should feed the release notes instead of PSR, the way `scriv github-release` did at `release.yml:60` before this PR.  Which way do you want to go?

.github/workflows/ci.yml
feanil: I think this is not needed.  `tox_uv` and the uv used to run CI use different python environments. I ran the CI sequence with PATH uv 0.12.18 against the locked 0.11.32 in the tox venv and `uv sync`, `uv lock --check` and `uv run tox -e quality` all passed with `uv.lock` untouched and no complaints as far as I could tell.

.github/workflows/release.yml
feanil: nit: this job only downloads an artifact and publishes it, so adding `contents: read` here would make it explicitly read only.

Makefile
feanil: `docs:` now inlines the five commands from `tox.ini`'s `[testenv:docs]`, so they become a second copy to keep in sync.  `uv run tox -e docs` keeps them in one place.  Was there a reason to inline them?

PR #389, opened 2026-08-31
chore: upgrade python requirements and fix isort invocation

pwnage101:
`isort`'s `--recursive` flag has been deprecated and a no-op for a long time. It's finally a fatal error, forcing us to remove it.

This also removes the "test_settings.py" arg which points to a file which doesn't exist anymore.

This also contains the result of `make upgrade`.

Note for reviewers: This does NOT need a version bump because it results in no change to the published package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ quality: ## check coding style with pycodestyle and pylint
 	mypy
 	pycodestyle openedx_filters  *.py
 	ruff check openedx_filters *.py
-	isort --check-only --diff --recursive test_utils openedx_filters *.py test_settings.py
+	isort --check-only --diff test_utils openedx_filters *.py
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -65,7 +65,7 @@ quality: ## check coding style with pycodestyle and pylint
 	mypy
 	pycodestyle openedx_filters  *.py
 	ruff check openedx_filters *.py
-	isort --check-only --diff --recursive test_utils openedx_filters *.py test_settings.py
+	isort --check-only --diff test_utils openedx_filters *.py
```
pwnage101, line 68: This is the only line I changed manually.

felipemontoya (approved):
Found this in a different repo I also maintain. This looks good

PR #390, opened 2026-09-03
feat: add SupportContactContextRequested filter

brobro10000:
ENT-11574

Adds a filter that lets pipeline steps enrich the support-contact tags list:
`SupportContactContextRequested` (`org.openedx.learning.support.contact.context.requested.v1`).

## Related PRs

* openedx/edx-enterprise — pipeline step: https://github.com/openedx/edx-enterprise/pull/2688
* openedx/openedx-platform — call-site swap, merges last: https://github.com/openedx/openedx-platform/pull/39076
* edx/edx-platform — sibling PR: https://github.com/edx/edx-platform/pull/455

## Testing

New: unit tests for the filter class in `openedx_filters/learning/tests/test_filters.py`,
matching the existing per-filter-class test pattern (asserts `run_filter` invokes the pipeline
and returns the expected value).

Locally verified: `pytest openedx_filters/learning/tests/test_filters.py` — 75 passed, 99% line
coverage on `filters.py`. `mypy` clean. `isort --check-only` clean.

## Changelog / version

Bumped `__version__` to `3.10.0` and added a `CHANGELOG.rst` entry, matching current practice
(see #391). Note: #391 is also unmerged and claims 3.10.0 — whichever of us merges second will
need a quick rebase to the next version slot; that's expected/normal, not a blocker.

### DIFF
--- a/openedx_filters/learning/filters.py
+++ b/openedx_filters/learning/filters.py
@@ -1936,3 +1936,41 @@ class CourseModePriceRequested(OpenEdxPublicFilter):
         """
         data = super().run_pipeline(user=user, course_mode_data=course_mode_data, price=price)
         return data["user"], data["course_mode_data"], data["price"]
+
+
+class SupportContactContextRequested(OpenEdxPublicFilter):
+    """
+    Filter used to enrich the support contact request context with custom tags.
+
+    Purpose:
+        This filter is triggered when a user submits a support contact request. Pipeline steps
+        can inspect the user to append custom tags to the tags list that will be associated
+        with the support ticket.
+
+    Filter Type:
+        org.openedx.learning.support.contact.context.requested.v1
+
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: lms/djangoapps/support/views/contact_us.py
+        - Function or Method: ContactUsView.get
+    """
+
+    filter_type = "org.openedx.learning.support.contact.context.requested.v1"
+
+    @classmethod
+    def run_filter(cls, tags: list[str], user: Any) -> tuple[list, Any]:
+        """
+        Process the tags list through the configured pipeline steps.
+
+        Arguments:
+            tags (list[str]): the list of tags to be associated with the support ticket.
+            user (User): the user submitting the support request.
+
+        Returns:
+            tuple[list, Any]:
+                - list: the (possibly modified) tags list.
+                - Any: the Django User object (unchanged).
+        """
+        data = super().run_pipeline(tags=tags, user=user)
+        return data["tags"], data["user"]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,17 @@ Unreleased
 ----------
 .. scriv-insert-here
 
+
+.. _changelog-3.10.0:
+
+[3.10.0] - 2026-09-08
+---------------------
+
+Added
+~~~~~
+
+* Added new ``SupportContactContextRequested`` filter
+
 .. _changelog-3.9.0:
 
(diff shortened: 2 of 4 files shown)


openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,40 @@ def run_filter(
...
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: lms/djangoapps/support/views/contact_us.py
+        - Function or Method: ContactUsView.get
+    """
+
+    filter_type = "org.openedx.learning.support.contact.context.requested.v1"
+
+    @classmethod
+    def run_filter(cls, tags: list, request: Any, user: Any) -> list:
```
pwnage101: filters should always return a tuple containing the same shape as the input arguments
```suggestion
    def run_filter(cls, tags: list, request: Any, user: Any) -> tuple[list, Any, Any]:
```

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,40 @@ def run_filter(
...
+        """
+        Process the tags list through the configured pipeline steps.
+
+        Arguments:
+            tags (list): the list of tags to be associated with the support ticket.
+            request (HttpRequest): the current HTTP request.
```
pwnage101: We almost never need to pass the request, which is available via `crum` within pipeline step implementations.
```suggestion
```

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,40 @@ def run_filter(
...
+            user (User): the user submitting the support request.
+
+        Returns:
+            list: the (possibly modified) tags list.
+        """
+        data = super().run_pipeline(tags=tags, request=request, user=user)
```
```suggestion
        data = super().run_pipeline(tags=tags, user=user)
```

openedx_filters/learning/filters.py
brobro10000: Done — removed `request` from the signature entirely (see the other thread) and switched to returning `tuple[list, Any]` (`tags, user`), matching the convention used by `AccountSettingsReadOnlyFieldsRequested`/`CourseModePriceRequested`. Pushed.

openedx_filters/learning/filters.py
brobro10000: Done — dropped `request` from `run_filter`. The edx-enterprise pipeline step now fetches it internally via `crum.get_current_request()` (edx-enterprise#2688), matching `courseware.py`/`dashboard.py`/`enrollment.py`/`logistration.py`/`course_modes.py`. Pushed.

openedx_filters/learning/filters.py
brobro10000: Applied, thanks.

openedx_filters/learning/filters.py
```diff
@@ -1936,3 +1936,41 @@ def run_filter(
...
+    Trigger:
+        - Repository: openedx/edx-platform
+        - Path: lms/djangoapps/support/views/contact_us.py
+        - Function or Method: ContactUsView.get
+    """
+
+    filter_type = "org.openedx.learning.support.contact.context.requested.v1"
+
+    @classmethod
+    def run_filter(cls, tags: list, user: Any) -> tuple[list, Any]:
```
pwnage101: Nit: you can clarify the list element type 
```suggestion
    def run_filter(cls, tags: list[str], user: Any) -> tuple[list, Any]:
```

pwnage101 (approved):

openedx_filters/learning/filters.py
brobro10000: Applied, thanks.

PR #393, opened 2026-09-08
feat: add SupportEnrollmentDataRequested filter

brobro10000:
ENT-11574

Companion PR to the already-open support-contact-tag PR set (openedx-filters#390,
edx-enterprise#2688, openedx-platform#39076, edx-platform#455) — this one covers the
other half of ENT-11574's acceptance criteria: enterprise enrollment data for the support
enrollment view.

Adds `SupportEnrollmentDataRequested` (`org.openedx.learning.support.enrollment.data.requested.v1`)
— lets pipeline steps inject enterprise enrollment data into the support enrollment lookup.

## Related PRs

* openedx/edx-enterprise — pipeline step: https://github.com/openedx/edx-enterprise/pull/2690
* openedx/openedx-platform — call-site swap, merges last: https://github.com/openedx/openedx-platform/pull/39083
* edx/edx-platform — sibling PR: https://github.com/edx/edx-platform/pull/462
* Sibling ticket-half already in review: openedx-filters#390, edx-enterprise#2688,
  openedx-platform#39076, edx-platform#455 (support-contact-tag)

## Testing

New: unit tests for the filter class in `openedx_filters/learning/tests/test_filters.py`,
matching the existing per-filter-class test pattern.

Locally verified: `pytest openedx_filters/learning/tests/test_filters.py` — 75 passed. `mypy`
clean. `isort --check-only` clean.

## Changelog / version

Bumped `__version__` to `3.11.0` (3.10.0 already claimed by the sibling contact-tag PR) and
added a `CHANGELOG.rst` entry, matching current practice.

PR #396, opened 2026-09-14
fix: change SupportEnrollmentDataRequested to accept/return full enrollments list

brobro10000:
## Description

Addresses [review feedback](https://github.com/edx/edx-platform/pull/462#discussion_r3972331858) ([and the follow-up](https://github.com/edx/edx-platform/pull/462#discussion_r3972354578)) on the sibling edx-platform PR #462: `SupportEnrollmentDataRequested`'s original shape was a "red flag" — it passed an empty placeholder dict to the filter, which meant pipeline steps could only ever bolt on one hard-coded key, and forced the platform view to manually zip the filter's results back onto each enrollment record via a for-loop afterward.

`run_filter` now takes the real `enrollments` list (already carrying every field the view has computed so far) and returns it back, so pipeline steps can augment any part of each enrollment record in place, not just append a fixed key.

`user` stays as an explicit argument here (unlike the sibling contact filter's fix) — this is a support tool where staff look up *other* users' enrollments, so `crum`'s current-request user would be the wrong value.

Since the original shape (#393) already merged and released, this is a new PR rather than an amendment.

## Companion PRs

- edx-enterprise: pipeline step update to match (forthcoming, same PR set)
- openedx-platform / edx-platform: call-site update deferred until this and the edx-enterprise companion PR are released, to avoid breaking their CI against the old pinned version

ENT-11574


openedx_filters/learning/filters.py
```diff
@@ -1993,18 +1994,18 @@ class SupportEnrollmentDataRequested(OpenEdxPublicFilter):
     filter_type = "org.openedx.learning.support.enrollment.data.requested.v1"
 
     @classmethod
-    def run_filter(cls, enrollment_data: dict, user: Any) -> tuple[dict, Any]:
+    def run_filter(cls, enrollments: list, user: Any) -> tuple[list, Any]:
```
pwnage101, line 1997: There's no reason to remove the "_data" suffix from the argument name.  Conventionally, that signifies the value represents a serialized object, typically a dict, rather than an ORM object or other semantic type.  I think the only change to the name should be to make it plural, and enhance the type: `enrollments_data: list[dict]`

openedx_filters/learning/filters.py
```diff
@@ -1993,18 +1994,18 @@ class SupportEnrollmentDataRequested(OpenEdxPublicFilter):
...
         """
-        Process the enrollment data dict through the configured pipeline steps.
+        Process the enrollments list through the configured pipeline steps.
 
         Arguments:
-            enrollment_data (dict): dict mapping course_id to list of enrollment records.
+            enrollments (list): list of enrollment record dicts, each keyed by course_id.
```
pwnage101: This should explain what keys to expect inside the serialized dict.  openedx-filters is the plugin-facing interface, treating the platform as a black box.  Plugin implementers (i.e. pipeline step implementers) should not necessarily need to spelunk platform code to understand what are the most essential keys available in the dict.

Therefore, at minimum you should document some of the most important dict keys, e.g. `course_id`, `course_start/end`, `enrollment_start/end`, `course_modes`, etc.  Maybe document it in the form of a sample payload.  Finally I'd recommend linking directly to the [serialization code](https://github.com/openedx/openedx-platform/blob/9214cb9/lms/djangoapps/support/views/enrollments.py#L113-L127) within this docstring for convenience.

pwnage101 (commented):
Otherwise, LGTM!

openedx_filters/learning/filters.py
brobro10000: Fixed — renamed back to `enrollments_data: list[dict]`, keeping the `_data` suffix convention. run_pipeline/return also use `enrollments_data` now.

openedx_filters/learning/filters.py
brobro10000: Fixed — the docstring now lists the expected keys (course_id, course_start, course_end, enrollment_start, enrollment_end, course_modes) and links to edx-platform's serialization code for the full shape.

brobro10000: Updated with real, verified keys instead of guessed ones — checked openedx.core.djangoapps.enrollments.api.get_enrollments's own docstring (which has a sample payload) plus what EnrollmentSupportListView.get adds before calling this filter (verified_price, verified_upgrade_deadline, verification_deadline, order_number, source_system, manual_enrollment), and linked to the live serialization code.

brobro10000: Restructured as a literal example payload (matching "sample payload") instead of prose — verified against get_enrollments's own doctest example plus what EnrollmentSupportListView.get adds before calling this filter (verified_price, verified_upgrade_deadline, verification_deadline, order_number, source_system, manual_enrollment).

pwnage101 (approved):

openedx_filters/learning/filters.py
```diff
@@ -1993,18 +1994,46 @@ class SupportEnrollmentDataRequested(OpenEdxPublicFilter):
     filter_type = "org.openedx.learning.support.enrollment.data.requested.v1"
 
     @classmethod
-    def run_filter(cls, enrollment_data: dict, user: Any) -> tuple[dict, Any]:
+    def run_filter(cls, enrollments_data: list[dict], user: Any) -> tuple[list[dict], Any]:
```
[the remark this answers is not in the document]
pwnage101: This has only been live for such a small amount of time that it's not warranted to version the filter.

